PR #2, opened 2022-06-15
Replaces the use of Makefiles with the Autotools

ilumsden:
This PR replaces the vanilla Makefiles with the use of the Autotools. As a result of this PR, DYAD now depends on the following:
* `autoconf`
* `automake`
* `libtool`
* `pkg_config` (to locate and load external packages, like Flux-Core)

Because a version number is required for the Autotools, this PR currently gives DYAD a placeholder version of 0.1.0. The "real" initial version (including development status, like alpha, beta, etc.) should probably be decided before merging this PR.

This PR also copies two files from [flux-core](https://github.com/flux-framework/flux-core) into DYAD:
* `autogen.sh` (a super simple shell script that runs libtoolize and autoreconf)
* `config/ax_cxx_compile_stdcxx.m4` (defines a macro for configure.ac that can check for a minimum required version of C++)

Finally, this PR adds a new `dyadrun` script. This script allows users to set DYAD's environment variables using command-line arguments (instead of having to remember all those env vars). Additionally, it will also run the specified program (by interpreting the first unrecognized, non-flag argument as the program name). So, for example, if you have the application `consumer` that uses the wrapper API, DYAD is managing the path `/tmp/app`, and DYAD's KVS namespace is `test`, you could run the following:
```bash
$ dyadrun -c -pre -path /tmp/app -n test consumer
```

If searching for this script in the repository, it is actually under the name `dyadrun.in`. When building DYAD, Automake will replace some template variables defined in `dyadrun.in` to create `dyadrun`. During install, the build system will install `dyadrun` into the `bindir`.

@JaeseungYeom one thing that will need to be done in the future for `dyadrun` is that the usage statement will need to be updated so that the flags corresponding to `DYAD_SHARED_STORAGE`, `DYAD_NUM_CPA_POINTS`, and `DYAD_SYNC_CHECK` are documented. Additionally, if any of the other flags' documentation is incorrect, please let me know, and I'll fix it.


JaeseungYeom:
I think you can rebase the commits to a smaller number of commits. I see you are using flux-core autogen.sh as the reference. I wonder if it makes sense to look into flux-sched instead because flux-sched requires flux-core installation, which is similar to dyad. Don't worry about `DYAD_NUM_CPA_POINTS`. Anything CPA related will be removed from the codebase.

ilumsden:
I can definitely rebase. Just wanted to get everything onto GitHub first. As for the `autogen.sh` stuff, I can definitely take a look at flux-sched, although I don't think it'll make much of a difference. `autogen.sh` is literally 2 lines of code (plus some `echo`s).

ilumsden:
@JaeseungYeom the only difference between the version of `autogen.sh` in flux-core and the version in flux-sched is that the flux-core version uses the `--force` flag with `autoreconf`. Otherwise, they're identical.

ilumsden:
@JaeseungYeom for the initial version info in `configure.ac`, we should be using SemVer 2.0.0 if we're sticking with the Flux project's release structure (https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_1.html#release-process).

Per SemVer 2.0.0, we should probably be using 0.Y.Z version numbers right now. These version numbers are meant for 100% stable public APIs. Although I don't think the APIs are going to change much, if at all, for C and C++, I personally think it would make sense to avoid locking the APIs in for the time being

JaeseungYeom (approved):
Looks good to me.

JaeseungYeom:
if you need to change anything for PerfFlowAspect, make that change, rebase and merge.

ilumsden:
[Finalize and merge Autotools PR](https://trello.com/c/UeRt523v/5-finalize-and-merge-autotools-pr)

JaeseungYeom:
please squash the commits into one.

JaeseungYeom:
All the commits except the PFA support should be in one commit. So, there should be two commits.

PR #3, opened 2022-06-16
Adds a License

ilumsden:
This is something minor, but important, that I thought of while talking with some members of GCLab.

This PR adds a License to DYAD. The specific license added is LGPL-3, the same license that appears to be used by every other repo in flux-framework.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,165 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+
+  This version of the GNU Lesser General Public License incorporates
+the terms and conditions of version 3 of the GNU General Public
+License, supplemented by the additional permissions listed below.
+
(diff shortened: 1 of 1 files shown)


garlick:
This might be a bit more complex since there are GPL headers on some of the source files and GPL cannot be included in a LGPL project.  Also you're missing the NOTICE.LLNS file, and should note the LLNL release number in the README.md like the other framework projects.

I assume it was decided early on that dyad was part of the flux framework and would fall under the existing license/release number.  Just be aware that originally flux was GPL, then we got a new number and switched to LGPL.  Part of switching meant excising any vendored source that had a license that was incompatible with LGPL.  We could just change the license on the files that we wrote. I think you have a choice of remaining GPL here if convenient or updating to the new release number and LGPL.

ilumsden:
Thanks for the info @garlick. To be honest, I had just added this because I thought it would be useful to have a license now that DYAD is open source. In hindsight, I probably should have discussed with @JaeseungYeom before making this PR.

Still, on my end, it's quite useful to know about these details. Thanks again.

JaeseungYeom:
Thank for letting us know the details. Yes, the plan was that dyad to be a part of flux framework. Dong created the placeholder repo a while ago. We are trying to populate the repo now. I do not see a reason not to follow Flux License. However, we will check.

PR #5, opened 2022-07-14
exp: add the barrier in the producer-consumer experiment script generation

JaeseungYeom:
In the script that generates job scripts for experimenting producer-consumer patterns, a barrier is used before starting any of producer and consumer such that they all can start at the same time regardless of the scheduler delay. This is to help obtain better measurements.


ilumsden (approved):
Looks good to me

PR #8, opened 2022-08-03
Build: automatically track the version based on git tag

JaeseungYeom:
Have `autoconfig` find DYAD version from git tag.


ilumsden (approved):
Looks good. Approved!

PR #9, opened 2022-08-03
Cleanup unmaintained/unused code

JaeseungYeom:
Clean up code no longer in use or unmaintained.
That includes codes written for collecting data to allow critical path analysis and simple timing measurements. We will rely on 3rd-party existing tools such as `PerfFlowAspect` instead.


ilumsden (approved):
Looks good. Approved!

PR #14, opened 2022-09-29
Creates a "core" library that implements all the shared functionality across APIs

ilumsden:
Previously, each API re-implements the main (or "core") functionality of DYAD, such as Flux KVS put/get operations and Flux RPC calls. This PR abstracts all this shared functionality into an underlying `libdyad_core.so` library with the following API:

```c
// Store all the data needed to use DYAD
struct dyad_ctx;
typedef struct dyad_ctx dyad_ctx_t;

// Store information fetched from the Flux KVS
struct dyad_kvs_response;
typedef struct dyad_kvs_response dyad_kvs_response_t;

// Initialize DYAD and create a "dyad_ctx_t" object
int dyad_init(bool debug, bool check, bool shared_storage,
        unsigned int key_depth, unsigned int key_bins, 
        const char *kvs_namespace, const char* prod_managed_path,
        const char *cons_managed_path, bool intercept, dyad_ctx_t **ctx);

// Save (or "commit") file metadata into the Flux KVS
int dyad_commit(dyad_ctx_t *ctx, const char *fname);

// Fetch file metadata from the Flux KVS
int dyad_fetch(dyad_ctx_t *ctx, const char* fname, 
        dyad_kvs_response_t **resp);

// Initiate a Flux RPC call and retrieve (or "pull") the file from the producer
int dyad_pull(dyad_ctx_t *ctx, const char* fname,
        dyad_kvs_response_t *kvs_data);

// Free the dyadd_kvs_response_t objects created by "dyad_fetch"
int dyad_free_kvs_response(dyad_kvs_response_t *resp);

// Finalize DYAD and deallocate the "dyad_ctx_t" object
int dyad_finalize(dyad_ctx_t *ctx);
```

The library also provides two convenience functions for producing and consuming data:

```c
// Wraps "dyad_commit"
int dyad_produce(dyad_ctx_t *ctx, const char *fname);
// Wraps "dyad_fetch" and "dyad_pull"
int dyad_consume(dyad_ctx_t *ctx, const char *fname);
```

Besides creating this library, this PR updates the C "sync" API and the C++ "stream" API to use the new "core" library.

Currently, to use the "core" library, the Automake files do the following:
* Link the API libraries against `libdyad_core.la` using libtool
  * When doing this, libtool automatically sets the `RPATH` of the API libraries to the path stored in the Autoconf variable `libdir`
* Install `libdyad_core.so` into `libdir`

This should mean that user code can use the API libraries without needing to manually link against `libdyad_core.so`. Alternatively, `libdyad_core.la` can be made a `noinst` target. However, this would prevent certain future decisions in the API libraries, such as making the C++ API a header-only library.

Finally, this PR makes some minor reorganization changes to the repo to make it easier for new developers or people outside the project to understand where code is. More specifically, it makes the following changes:
* Renames `common` to `utils`
* Renames `lib` to `stream`
* Moves the URPC library into its own `urpc` directory


src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    return 0;
+}
+
+int dyad_init(bool debug, bool check, bool shared_storage,
+        unsigned int key_depth, unsigned int key_bins,
+        const char *kvs_namespace, const char *prod_managed_path,
+        const char *cons_managed_path, bool intercept, dyad_ctx_t **ctx)
+{
+    printf("In dyad_init\n");
+    if (*ctx != NULL)
```
JaeseungYeom, line 85: Also need to check if ctx != NULL first.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    (*ctx)->key_bins = key_bins;
+    (*ctx)->intercept = intercept;
+    printf("Setting KVS namespace\n");
+    if (kvs_namespace == NULL)
+    {
+        fprintf (stderr, "No KVS namespace provided!\n");
+        // TODO see if we want a different return val
+        return DYAD_NOCTX;
+    }
+    (*ctx)->kvs_namespace = (char*) malloc(strlen(kvs_namespace)+1);
```
JaeseungYeom, line 120: Need to initialize or set `(*ctx)->kvs_namespace[strlen(kvs_namespace)] = '\0'`

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+            fprintf (stderr, "Could not allocate buffer for Producer managed path!\n");
+            free((*ctx)->kvs_namespace);
+            free(*ctx);
+            *ctx = NULL;
+            return DYAD_NOCTX;
+        }
+        strncpy(
+                (*ctx)->prod_managed_path,
+                prod_managed_path,
+                strlen(prod_managed_path)+1
```
JaeseungYeom, line 152: There is no error with this way of handling null-termination. However, imagine the case we would need to support string of wide characters, in which case the null-termination would require more than 1 byte. I think it is safer to do null-termination in more explicit fashion either via the character buffer initialization or just writing a null-termination character at the end.

src/core/dyad_core.c
JaeseungYeom, line 120: Since kvs_namespace is const array, you can store the length in a const variable and reuse whenever you need call `strlen()` again. `strlen()` will walk through the entire buffer until it reaches the null-terminating character.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        strncpy(
+                (*ctx)->cons_managed_path,
+                cons_managed_path,
+                strlen(cons_managed_path)+1
+            );
+    }
+    if ((*ctx)->prod_managed_path == NULL && (*ctx)->cons_managed_path == NULL)
+    {
+        fprintf (stderr, "Warning: DYAD instance is neither producer nor consumer!\n");
+    }
```
JaeseungYeom, line 181: This check could have happened at the beginning, and skip the initialization other than setting `(*ctx)->intercept = false`

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    printf("Opening Flux\n");
+    (*ctx)->reenter = true;
+    (*ctx)->h = flux_open(NULL, 0);
+    if ((*ctx)->h == NULL)
+    {
+        fprintf (stderr, "Could not open Flux handle!\n");
+        return DYAD_FLUXFAIL;
+    }
```
JaeseungYeom, line 189: This check also should happen at the top.

src/core/dyad_err.h
```diff
@@ -0,0 +1,34 @@
+#ifndef __DYAD_CORE_DYAD_ERR_H__
+#define __DYAD_CORE_DYAD_ERR_H__
+
+enum dyad_core_err_codes {
```
JaeseungYeom, line 4: rename to return code instead of error code

src/core/dyad_err.h
```diff
@@ -0,0 +1,34 @@
...
+    DYAD_BADFETCH = -5, // Flux KVS commit didn't work
+    DYAD_BADRESPONSE = -6, // Cannot create/populate a DYAD response
+    DYAD_BADRPC = -7, // Flux RPC pack or get didn't work
+    DYAD_BADFIO = -8, // File I/O failed
+    DYAD_BADMANAGEDPATH = -9, // Cons or Prod Manged Path is bad
+};
+
+typedef enum dyad_core_err_codes dyad_core_err_t;
+
+#define DYAD_IS_ERROR(code) (code < 0)
```
JaeseungYeom, line 19: `((code) < 0)`

src/core/dyad_open_dlwrap.c
```diff
@@ -0,0 +1,135 @@
+#include "dyad_open_dlwrap.h"
```
JaeseungYeom, line 1: The file name can change to io_interceptor or something else because  it is not only for open but also for close

src/core/dyad_core.h
```diff
@@ -0,0 +1,192 @@
...
+    bool initialized; // if true, DYAD is initialized
+    bool shared_storage; // if true, the managed path is shared
+    bool sync_started; // TODO
+    unsigned int key_depth; // Depth of bins for the Flux KVS
+    unsigned int key_bins; // Number of bins for the Flux KVS
+    uint32_t rank; // Flux rank for DYAD
+    char *kvs_namespace; // Flux KVS namespace for DYAD
+    char *prod_managed_path; // producer path managed by DYAD
+    char *cons_managed_path; // consumer path managed by DYAD
+    bool intercept; // if true, intercepts (f)open and (f)close
```
JaeseungYeom, line 51: I think reenter was serving the same purpose in the original code, but I might be wrong.

src/core/dyad_core.h
```diff
@@ -0,0 +1,192 @@
...
+{
+#endif
+
+/**
+ * @struct dyad_ctx
+ */
+struct dyad_ctx {
+    flux_t *h; // the Flux handle for DYAD
+    bool debug; // if true, perform debug logging
+    bool check; // if true, perform some check logging
```
JaeseungYeom, line 40: Not sure what the purpose of check flag is for

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    false,
+    true,
+    false,
+    false,
+    3u,
+    1024u,
+    0u,
+    NULL,
+    NULL
+};
```
JaeseungYeom, line 24: missing the initial value for intercept

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
+#include "dyad_core.h"
+#include "dyad_err.h"
+#include "utils.h"
+#include "murmur3.h"
```
JaeseungYeom, line 4: I guess these are not installed headers. So, it is ok to be without the subfoler location of headers. If it was cmake, it would have been better to match the header files with the actual location of headers. e.g., "common/murmur3.h"

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+
+#include <libgen.h>
+#include <limits.h>
```
JaeseungYeom, line 7: `#ifdef __cplusplus` for cstring amd climits

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+}
+
+int gen_path_key (const char* str, char* path_key, const size_t len,
+        const uint32_t depth, const uint32_t width)
+{
+    static const uint32_t seeds [10]
+        = {104677, 104681, 104683, 104693, 104701,
+            104707, 104711, 104717, 104723, 104729};
+
+    uint32_t seed = 57;
```
JaeseungYeom, line 50: 57u

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        return DYAD_BADMANAGEDPATH;
+    }
+    return DYAD_OK;
+}
+
+int gen_path_key (const char* str, char* path_key, const size_t len,
+        const uint32_t depth, const uint32_t width)
+{
+    static const uint32_t seeds [10]
+        = {104677, 104681, 104683, 104693, 104701,
```
JaeseungYeom, line 47: add 'u' at the end of each number

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        free(ctx->prod_managed_path);
+        ctx->prod_managed_path = NULL;
+    }
+    if (ctx->cons_managed_path != NULL)
+    {
+        free(ctx->cons_managed_path);
+        ctx->cons_managed_path = NULL;
+    }
+    free(ctx);
+    ctx = NULL;
```
JaeseungYeom, line 584: Do we want to use `**ctx` to make sure `ctx = NULL` to have effect, or is this fine?

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+
+int dyad_free_kvs_response(dyad_kvs_response_t *resp)
+{
+    if (resp == NULL)
+    {
+        return 0;
+    }
+    free(resp->fpath);
+    free(resp);
+    resp = NULL;
```
JaeseungYeom, line 553: Do we want to use **resp to make sure resp = NULL to have effect, or is this fine?

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    }
+
+    of = fopen(file_path, "w");
+    if (of == NULL)
+    {
+        DYAD_LOG_ERR(ctx, "Cannot open file %s\n", file_path);
+        rc = DYAD_BADFIO;
+        goto pull_done;
+    }
+    size_t written_len = fwrite(file_data, sizeof(char), (size_t) file_len, of);
```
JaeseungYeom, line 522: const size_t

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    {
+        setenv(DYAD_CHECK_ENV, "ok", 1);
+    }
+    return rc;
+}
+
+#if DYAD_PERFFLOW
+__attribute__((annotate("@critical_path()")))
+#endif
+int dyad_kvs_lookup(dyad_ctx_t *ctx, const char* kvs_topic, uint32_t *owner_rank)
```
JaeseungYeom, line 348: can ctx be const?

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    }
+    if (rc < 0)
+    {
+        DYAD_LOG_ERR(ctx, "Could not unpack owner's rank from KVS response\n");
+        return DYAD_BADFETCH;
+    }
+    return DYAD_OK;
+}
+
+int dyad_fetch(dyad_ctx_t *ctx, const char* fname,
```
JaeseungYeom, line 376: can ctx be const?

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    3u,
+    1024u,
+    0u,
+    NULL,
+    NULL
+};
+
+int validate_context(dyad_ctx_t *ctx, bool check_cons, bool check_prod)
+{
+    if (!ctx || !ctx->h)
```
JaeseungYeom, line 28: Styling-wise, the opening brace better be at the end of the same line with `if` unless the line becomes too long, or the block has too many lines.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+#include "utils.h"
+#include "murmur3.h"
+
+#include <libgen.h>
+#include <limits.h>
+#include <string.h>
+#include <unistd.h>
+
+const struct dyad_ctx dyad_ctx_default = {
+    NULL,
```
JaeseungYeom, line 12: add inline comment after the comma to show what variable the value is for

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    printf("Initial publication setup\n");
+    const char *prod_managed_path = ctx->prod_managed_path;
+    flux_kvs_txn_t *txn = NULL;
+    flux_future_t *f = NULL;
+    const size_t topic_len = PATH_MAX;
+    char topic[topic_len+1];
+    printf("Creating KVS key\n");
+    memset(topic, '\0', topic_len+1);
+    gen_path_key(
+            upath,
```
JaeseungYeom, line 279: In case that there are too many arguments, and need to line-break to avoid an extremely long line, those arguments should be aligned with the next character to the opening parenthesis whenever possible.

src/core/dyad_err.h
```diff
@@ -0,0 +1,34 @@
+#ifndef __DYAD_CORE_DYAD_ERR_H__
+#define __DYAD_CORE_DYAD_ERR_H__
+
+enum dyad_core_err_codes {
+    DYAD_OK = 0, // Operation worked correctly
```
JaeseungYeom, line 5: add RC_.  e.g., RC_DYAD_OK

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        f = NULL;
+    }
+    if (rc < 0)
+    {
+        DYAD_LOG_ERR(ctx, "Cannot get file back from plugin via RPC!\n");
+        return DYAD_BADRPC;
+    }
+    return DYAD_OK;
+}
+
```
JaeseungYeom, line 471: doxygen documentation needed

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    (*resp)->fpath = malloc(strlen(upath)+1);
+    strncpy(
+            (*resp)->fpath,
+            upath,
+            strlen(upath)+1
+           );
+    (*resp)->owner_rank = owner_rank;
+    return DYAD_OK;
+}
+
```
JaeseungYeom, line 436: doxygen documentation needed for the function below

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        f = NULL;
+    }
+    if (rc < 0)
+    {
+        DYAD_LOG_ERR(ctx, "Could not unpack owner's rank from KVS response\n");
+        return DYAD_BADFETCH;
+    }
+    return DYAD_OK;
+}
+
```
JaeseungYeom, line 375: doxygen documentation needed for the function below

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    ctx->reenter = true;
+
+commit_done:
+    if (rc == DYAD_OK && (ctx && ctx->check))
+    {
+        setenv(DYAD_CHECK_ENV, "ok", 1);
+    }
+    return rc;
+}
+
```
JaeseungYeom, line 344: doxygen documentation needed for the function below

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    printf("Calling dyad_kvs_commit\n");
+    int ret = dyad_kvs_commit(ctx, txn, upath, &f);
+    if (ret < 0)
+    {
+        DYAD_LOG_ERR(ctx, "dyad_kvs_commit failed!\n");
+        return ret;
+    }
+    return DYAD_OK;
+}
+
```
JaeseungYeom, line 313: doxygen documentation needed for the function below

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+#if DYAD_PERFFLOW
+__attribute__((annotate("@critical_path()")))
+#endif
+int publish_via_flux(dyad_ctx_t* ctx, const char *upath)
+{
+    printf("Initial publication setup\n");
+    const char *prod_managed_path = ctx->prod_managed_path;
```
JaeseungYeom, line 271: I suspect that this line would cause a compiler warning with certain C compilers. There are some that expects the variable initialization with a value that is known at compile time. So, to avoid such a warning, the pointer can be set to NULL at variable definition, and then later set to the value after all the local variables are initialized.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        return DYAD_FLUXFAIL;
+    }
+    printf("Pack producer rank into transaction\n");
+    if (flux_kvs_txn_pack(txn, 0, topic, "i", ctx->rank) < 0)
+    {
+        DYAD_LOG_ERR(ctx, "Could not pack Flux KVS transaction\n");
+        return DYAD_FLUXFAIL;
+    }
+    printf("Calling dyad_kvs_commit\n");
+    int ret = dyad_kvs_commit(ctx, txn, upath, &f);
```
JaeseungYeom, line 305: Same issue with compiler warning regarding variable initialization.

src/core/dyad_core.h
```diff
@@ -0,0 +1,192 @@
...
+ * @brief Commits (i.e., writes) the file's metadata to the Flux KVS
+ * @param[in] ctx    the DYAD context for the operation
+ * @param[in] fname  the name of the file being committed
+ *
+ * @return An integer error code (values TBD)
+ */
+#if DYAD_PERFFLOW
+__attribute__((annotate("@critical_path()")))
+#endif
+int dyad_commit(dyad_ctx_t *ctx, const char *fname);
```
JaeseungYeom, line 137: We should not have this function in this header. It should be moved to dyad_core.c and marked as static

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    *f = NULL;
+    printf("Destroy transaction on KVS\n");
+    flux_kvs_txn_destroy(txn);
+    return DYAD_OK;
+}
+
+#if DYAD_PERFFLOW
+__attribute__((annotate("@critical_path()")))
+#endif
+int publish_via_flux(dyad_ctx_t* ctx, const char *upath)
```
JaeseungYeom, line 268: This function needs to be static, and the symbol should not be visible to other compilation unit.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        DYAD_LOG_ERR (ctx, "dyad_free_kvs_response failed!\n");
+        return rc;
+    };
+    return DYAD_OK;
+}
+
+#if DYAD_PERFFLOW
+__attribute__((annotate("@critical_path()")))
+#endif
+int dyad_kvs_commit(dyad_ctx_t *ctx, flux_kvs_txn_t *txn,
```
JaeseungYeom, line 244: This function also needs to be static.

JaeseungYeom (commented):
- I am not seeing the check with `reenter`. Make sure that the behavior is consistent with the original code. It serves an important purpose.
- Add `__restrict__` keyword whenever you see a non-const pointer in the argument list of a function definition that also contains more pointer arguments. Use it even if a pointer argument is const in the presence of other non-const pointer argument or for possible future inclusion of such a pointer argument.
- Use `static` keyword to avoid exposing functions that have no business externally.
- Avoid defining a new local variable in the middle of function body in C code. Put them all at the beginning of the function and initialize them with values known at compile time. This is to avoid compiler warning with somewhat old C compilers. This does not apply to C++.
- Style suggestions for opening braces and line-breaking. You also need to run clang-format to be compliant.
- Need some clarification of (new) context variables: check, intercept
- Rename the enum symbols for return codes by appending `DYAD_RC_` or `RC_DYAD_`
- See if you can change some of non-const pointer arguments to const.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    NULL
+};
+
+int validate_context(dyad_ctx_t *ctx, bool check_cons, bool check_prod)
+{
+    if (!ctx || !ctx->h)
+    {
+        return DYAD_NOCTX;
+    }
+    if (check_cons && (ctx->cons_managed_path == NULL || strlen(ctx->cons_managed_path) == 0))
```
JaeseungYeom, line 32: Do it only once at the initialization if possible. If not, it is worth storing this result of this check and reuse it. `strlen()` will read every character in the string.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    {
+        DYAD_LOG_ERR(ctx, "dyad_kvs_commit failed!\n");
+        return ret;
+    }
+    return DYAD_OK;
+}
+
+int dyad_commit(dyad_ctx_t *ctx, const char *fname)
+{
+    int rc = validate_context(ctx, false, true);
```
JaeseungYeom, line 316: This is redundant. There is a same check at line 205.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    }
+    rc = dyad_kvs_lookup(ctx, topic, &owner_rank);
+    if (rc != DYAD_OK)
+    {
+        DYAD_LOG_ERR(ctx, "dyad_kvs_lookup failed!\n");
+        return rc;
+    }
+    if (ctx->shared_storage || (owner_rank == ctx->rank))
+    {
+        return 0;
```
JaeseungYeom, line 419: 0 should DYAD_OK

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+            ctx->key_depth,
+            ctx->key_bins
+            );
+    if (ctx->h == NULL)
+    {
+        fprintf (stderr, "No Flux handle found in dyad_fetch!\n");
+        return DYAD_NOCTX;
+    }
+    rc = dyad_kvs_lookup(ctx, topic, &owner_rank);
+    if (rc != DYAD_OK)
```
JaeseungYeom, line 412: `rc` should be checked against 0  even though `DYAD_OK` coincidentally has the same value.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    {
+        DYAD_LOG_ERR(ctx, "dyad_kvs_lookup failed!\n");
+        return rc;
```
JaeseungYeom, line 415: The return of this function should not be mixed. It should be either the return code of flux functions or the dyad return code.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    if (ctx == NULL)
+    {
+        fprintf (stderr, "DYAD context not provided to dyad_fetch\n");
+        return DYAD_NOCTX;
+    }
+    char upath[PATH_MAX] = {'\0'};
+    if (!cmp_canonical_path_prefix(ctx->cons_managed_path, fname, upath, PATH_MAX))
+    {
+        DYAD_LOG_INFO (ctx, "%s is not in the Consumer's managed path\n", fname);
+        return 0;
```
JaeseungYeom, line 393: This should be `DYAD_OK`

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    if (DYAD_IS_ERROR(rc))
+    {
+        return rc;
+    }
+    dyad_kvs_response_t *resp = NULL;
+    rc = dyad_fetch(ctx, fname, &resp);
+    if (rc != 0)
+    {
+        DYAD_LOG_ERR (ctx, "dyad_fetch failed!\n");
+        if (resp != NULL)
```
JaeseungYeom, line 225: if `rc != 0`, `resp` should be always be NULL.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+int dyad_consume(dyad_ctx_t *ctx, const char *fname)
+{
+    int rc = validate_context(ctx, true, false);
+    if (DYAD_IS_ERROR(rc))
+    {
+        return rc;
+    }
+    dyad_kvs_response_t *resp = NULL;
+    rc = dyad_fetch(ctx, fname, &resp);
+    if (rc != 0)
```
JaeseungYeom, line 222: It should be `rc != DYAD_OK`

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+        DYAD_LOG_ERR(ctx, "Could not unpack owner's rank from KVS response\n");
+        return DYAD_BADFETCH;
+    }
+    return DYAD_OK;
+}
+
+int dyad_fetch(dyad_ctx_t *ctx, const char* fname,
+        dyad_kvs_response_t **resp)
+{
+    int rc = validate_context(ctx, true, false);
```
JaeseungYeom, line 379: This is a redundant check. `dyad_consume()` already runs this check before calling this function.

src/core/dyad_core.c
```diff
@@ -0,0 +1,622 @@
...
+    if (rc != 0)
+    {
+        DYAD_LOG_ERR (ctx, "dyad_fetch failed!\n");
+        if (resp != NULL)
+        {
+            dyad_free_kvs_response(resp);
+        }
+        return rc;
+    }
+    rc = dyad_pull(ctx, fname, resp);
```
JaeseungYeom, line 231: `dyad_fetch()` can finish successfully. but resp could still be NULL. `dyad_pull` should be skipped.

src/stream/dyad_stream_core.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    IPRINTF (m_ctx, "DYAD_SYNC CLOSE: exists sync (\"%s\").\n", path);
+    return true;
+}
+
+} // end of namespace dyad
+
+/*
+ * vi: ts=4 sw=4 expandtab
+ * y
```
JaeseungYeom, line 238: 'y' should be removed?

src/stream/dyad_stream_core.cpp
```diff
@@ -0,0 +1,240 @@
...
+        return true;
+    }
+
+    int rc = 0;
+
+    if (! is_dyad_producer ()) {
+        return true;
+    }
+
+    rc = dyad_produce(m_ctx, path);
```
JaeseungYeom, line 223: This is ok. However, the `reenter` set/unset in the `dyad_produce()` is not relevant here as there is no interception of symbol.

src/core/dyad_open_dlwrap.c
ilumsden: Actually, I've phased this out in favor of `reenter`. I'm keeping the files around until I know everything is working, just to be safe. Assuming things work as expected, these files will be deleted.

src/core/dyad_core.h
ilumsden: Similar to the `dyad_open_dlwrap.h` stuff, the `intercept` flag has been phased out in favor of `reenter`. I'll remove it.

src/core/dyad_core.h
ilumsden: I'm not positive either. It's leftover from the definition of the context in `wrapper.h` (in the `main` branch). If it's not necessary, I'll get rid of it.

src/core/dyad_core.c
ilumsden: Oh, I didn't even think about the fact that `ctx = NULL` would only affect the local variable. I'll switch it to `**ctx`.

src/core/dyad_core.c
ilumsden: I'll change to `**resp`

src/core/dyad_core.c
ilumsden: There's two reasons I removed `__restrict__`:

1. It's not part of the C or C++ standards. I tried to minimize the amount of non-standard syntax in core to avoid any possible issues with using it in API libraries (e.g., stream and sync)
2. Although `__restrict__` is pretty well supported by most C compilers, most of what I'm seeing online says that support for `__restrict__` is much more iffy for C++ compilers. Since these functions are meant to be called from C++ (and possibly other languages in the future), I wanted to avoid `__restrict__` in the internal API for core.

That being said, functions like `dyad_kvs_lookup` that are not defined in the header could probably use `__restrict__`. I'll look into it for those.

src/core/dyad_core.c
ilumsden: I'm not worrying too much about style at this point. My plan is to get everything working and then run `clang-format` to correct style.

src/core/dyad_core.c
ilumsden: @JaeseungYeom if I were to add the restrict feature, would it be better to use `__restrict__` (i.e., compiler extension) or `restrict` (i.e., C99 keyword). Since our code is C99 (a requirement we inherit from Flux), we could just used the keyword, which should be supported by all C99-compliant compilers.

src/core/dyad_core.c
ilumsden: They are not mixed here. `dyad_kvs_lookup` returns a DYAD return code. That code is stored in `rc` before being used in this conditional.

src/core/dyad_core.c
ilumsden: I've done this for `kvs_namespace`. Would we like to do this same for the managed paths?

src/core/dyad_core.c
ilumsden: I've replaced this conditional (along with any similar ones I could find) with `DYAD_IS_ERROR`

src/core/dyad_core.c
ilumsden: I've added a conditional just before `dyad_pull` that will check for this.

src/core/dyad_core.c
JaeseungYeom: > I've done this for `kvs_namespace`. Would we like to do this same for the managed paths?

Yes

src/core/dyad_core.c
JaeseungYeom: > They are not mixed here. `dyad_kvs_lookup` returns a DYAD return code. That code is stored in `rc` before being used in this conditional.
I see that now.

ilumsden:
@JaeseungYeom I've fixed most of the comments you've made. The remaining things that are not yet addressed are:
* This comment: "I think there is a bug in consumer code regarding dyad_pull() not being skipped when it should be."
  * Where is this bug? I'm not entirely sure what you are referring to here.
* Stuff about `__restrict__`
  * I responded to the relevant comment above asking about whether I should use `__restrict__` (compiler extension) or `restrict` (C99 keyword)
* Style fixes
  * I intend to fix this using `clang-format` once everything is working
* Renaming the error code types and macros
  * I was trying to make something similar to what UCX does. To that end, maybe a type name like `dyad_status_t` would be better? If you're curious what I'm specifically looking at, the relevant UCX code is [here](https://github.com/openucx/ucx/blob/master/src/ucs/type/status.h#L45)
* Figuring out which pointers can be converted to `const`
* Function inlining
  * Since I'm not positive if inline is compatible with PerfFlow Aspect (PFA), I think I'll manually inline (with the `inline` keyword) when PFA is not enabled.

I've resolved all the comments above that I have either fixed/addressed. The only ones that are still open are ones that either haven't been addressed or ones that I have responded to.

src/core/dyad_core.c
```diff
@@ -0,0 +1,781 @@
...
+static inline int publish_via_flux(dyad_ctx_t* ctx, const char *upath)
+#endif
+{
+    printf("Initial publication setup\n");
+    int ret = 0;
+    const char* prod_managed_path = NULL;
+    flux_kvs_txn_t *txn = NULL;
+    flux_future_t *f = NULL;
+    const size_t topic_len = PATH_MAX;
+    char topic[topic_len+1] = {'\0'};
```
JaeseungYeom: `topic[topic_len+1]` is problematic with some C compilers.
It should be `topic[PATH_MAX+1]`

src/core/dyad_core.c
```diff
@@ -0,0 +1,781 @@
...
+#else
+static inline int dyad_fetch(dyad_ctx_t *ctx, const char* fname,
+        dyad_kvs_response_t **resp)
+#endif
+{
+    int rc = 0;
+    char upath[PATH_MAX] = {'\0'};
+    uint32_t owner_rank = 0;
+    const size_t topic_len = PATH_MAX;
+    char topic[topic_len+1] = {'\0'};
```
JaeseungYeom: It should be topic[PATH_MAX+1]

src/core/dyad_core.c
```diff
@@ -0,0 +1,781 @@
...
+    {
+        DYAD_LOG_ERR(ctx, "Could not unpack owner's rank from KVS response\n");
+        return DYAD_BADFETCH;
+    }
+    return DYAD_OK;
+}
+
+#if DYAD_PERFFLOW
+__attribute__((annotate("@critical_path()")))
+static int dyad_fetch(dyad_ctx_t *ctx, const char* fname,
```
JaeseungYeom, line 276: `dyad_ctx_t *ctx` -> `const dyad_ctx_t *ctx`

src/core/dyad_core.c
```diff
@@ -0,0 +1,781 @@
...
+        dyad_kvs_response_t **resp)
+#else
+static inline int dyad_fetch(dyad_ctx_t *ctx, const char* fname,
```
JaeseungYeom, line 279: `dyad_ctx_t *ctx` -> `const dyad_ctx_t *ctx`

JaeseungYeom: same for validate_context, dyad_kvs_commit, publish_via_flux, dyad_kvs_lookup, dyad_rpc_get, and dyad_pull

src/core/dyad_core.c
```diff
@@ -0,0 +1,788 @@
...
+        return DYAD_OK;
+    }
+    // Call dyad_pull to fetch the data from the producer's
+    // Flux broker
+    rc = dyad_pull(ctx, fname, resp);
+    // Set reenter to true to allow additional intercepting
+    ctx->reenter = true;
+    // Regardless if there was an error in dyad_pull,
+    // free the KVS response object
+    dyad_free_kvs_response(&resp);
```
JaeseungYeom, line 711: I think this function can just be a part of `dyad_pull()`

src/core/dyad_core.c
ilumsden: Done

src/core/dyad_core.c
ilumsden: Resolved

src/core/dyad_core.c
ilumsden: Resolved

src/core/dyad_core.c
ilumsden: Resolved

ilumsden:
@JaeseungYeom this PR is almost done. The only comment you made above that I haven't yet resolved is the comment about renaming the error/return codes and the associated type. Otherwise, everything you've mentioned above has been resolved. I have also tested this PR using the ECP demo workflow, and it is working as intended.

At this point, I am going to open the PR as "ready for review". We can discuss the return code naming later this week. Then, all that will be left is any final reviewing you want to do.

src/core/dyad_core.c
```diff
@@ -0,0 +1,706 @@
...
+    const dyad_ctx_t* restrict ctx,
+    const char* restrict fname,
+    dyad_kvs_response_t** restrict resp)
+#else
+static inline dyad_rc_t dyad_fetch (const dyad_ctx_t* restrict ctx,
+                              const char* restrict fname,
+                              dyad_kvs_response_t** restrict resp)
+#endif
+{
+    dyad_rc_t rc = 0;
```
JaeseungYeom, line 244: DYAD_RC_OK instead of 0

JaeseungYeom (approved):

PR #15, opened 2022-12-05
Adding a use case example 'shuffle'

JaeseungYeom:
Use case demonstration for input file shuffling with deep learning training.
DYAD needs to be updated to leverage `fcntl()` to dynamically determine the mode of each individual open file, and handle it accordingly. This code is independent of DYAD and Flux. However, it reuses some of utility functions in DYAD implementation. The source files of those are symbolically linked to avoid duplication.


ilumsden:
What's left  to do:
- Add script(s) to run with DYAD
- Integrate into testing

JaeseungYeom:
Merging this to help the development of eviction service

Issue #10, opened 2022-08-03
Dynamically determine the mode between producer and consumer


JaeseungYeom:
Currently, DYAD requires environment variables to distinguish between producer and consumer. This should change such that it can switch behavior based on how the file handled is open (e.g., write mode or read mode)
`fcntl()` can be useful to achieve this.

JaeseungYeom:
PR#16

PR #16, opened 2022-12-07
Wrapper: automatically detect if the file access sync need

JaeseungYeom:
DYAD currently detects the synchronization need by environment variable.
Users need to set appropriate variable to tell DYAD I/O interceptor which file access to synchronize. However, this has a limitation as a process can open multiple files with different modes. A consumer can be a producer at the same time, and vice versa. This PR removes the need for environment variable. For a file about to be closed, it utilizes the system call `fcntl()` to detect if a file has been open in the write mode. For a file about to be opened, the mode argument can be checked easily. This addresses issue #10

### DIFF
--- a/src/wrapper/wrapper.c
+++ b/src/wrapper/wrapper.c
@@ -721,8 +721,9 @@ real_call:; // semicolon here to avoid the error
 
     int wronly = is_wronly (fd);
 
-    if (wronly == -1)
+    if (wronly == -1) {
         DPRINTF ("Failed to check the mode of the file with fcntl: %s\n", strerror (errno));
+    }
 
     if (to_sync && wronly == 1) {
         rc = func_ptr (fd);
@@ -797,8 +798,9 @@ real_call:;
 
     int wronly = is_wronly (fd);
 
-    if (wronly == -1)
+    if (wronly == -1) {
         DPRINTF ("Failed to check the mode of the file with fcntl: %s\n", strerror (errno));
+    }
 
 
     if (to_sync && wronly == 1) {
--- a/src/wrapper/wrapper.c
+++ b/src/wrapper/wrapper.c
@@ -90,6 +90,11 @@ int sync_directory (const char* path);
  *                                                                           *
  *****************************************************************************/
 
+static inline bool is_wronly (int fd)
+{
+  return (O_WRONLY == fcntl(fd, F_GETFL) & O_ACCMODE);
+}
+
 static inline bool is_dyad_producer ()
 {
     char *e = NULL;
@@ -390,11 +395,9 @@ static int dyad_open_sync (const char* __restrict__ path,
                            const char* __restrict__ user_path)
 {
     int rc = 0;
-    if (is_dyad_consumer ()) {
-        ctx->reenter = false;
-        rc = subscribe_via_flux (dyad_path, user_path);
-        ctx->reenter = true;
-    }
+    ctx->reenter = false;
+    rc = subscribe_via_flux (dyad_path, user_path);
+    ctx->reenter = true;
     return rc;
 }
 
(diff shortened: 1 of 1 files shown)


src/wrapper/wrapper.c
```diff
@@ -90,6 +90,11 @@ int sync_directory (const char* path);
  *                                                                           *
  *****************************************************************************/
 
+static inline bool is_wronly (int fd)
+{
+  return (O_WRONLY == fcntl(fd, F_GETFL) & O_ACCMODE);
```
ilumsden, line 95: @JaeseungYeom this function currently won't work as expected. In C/C++, bitwise and (&) has a lower operator precedence than equality (==). So, right now, this line will compare the output of `fcntl` with `O_WRONLY` and then perform a logical and with `O_ACCMODE`. You need to put `fcntl(fd, F_GETFL) & O_ACCMODE` in parentheses to get the expected behavior.

ilumsden (dismissed):
@JaeseungYeom looks really good. Only thing that needs to change is the comment I made about order of operations with the output of `fcntl`. Once that change is made, I'll test this with the demo, and, if it works, I'll approve.

src/wrapper/wrapper.c
garlick: If I may insert a drive by comment also:

Suggestion: check for fcntl failure (-1) before the bitwise stuff and return false or whatever.  Even if an error is unlikely/impossible/irrelevant in the contexts where this function is called, not checking for the error will set off red flags for some people.

JaeseungYeom: > If I may insert a drive by comment also:
> 
> Suggestion: check for fcntl failure (-1) before the bitwise stuff and return false or whatever. Even if an error is unlikely/impossible/irrelevant in the contexts where this function is called, not checking for the error will set off red flags for some people.

An excellent suggestion. Thank you!

JaeseungYeom: > @JaeseungYeom this function currently won't work as expected. In C/C++, bitwise and (&) has a lower operator precedence than equality (==). So, right now, this line will compare the output of `fcntl` with `O_WRONLY` and then perform a logical and with `O_ACCMODE`. You need to put `fcntl(fd, F_GETFL) & O_ACCMODE` in parentheses to get the expected behavior.

Can you fix it and address Jim's comment below?

ilumsden: @JaeseungYeom just pushed the fix. The `is_wronly` function will now return an integer. It will return -1 on `fcntl` error, 0 if the file descriptor is not write-only, and 1 if the file descriptor is write-only. I've also updated the spots where `is_wronly` is used so that, if an error occurs, a message will be printed with `DPRINTF`, and the actual file I/O call will be invoked (as opposed to calling all the DYAD-specific stuff).

I'm working on testing right now.

ilumsden:
@JaeseungYeom I've tweaked your `is_wronly` function to check for `fcntl` errors (as @garlick recommended). I've tested this with my DYAD demo for the ECP tutorial, and it works. Can you take a look to make sure it looks good before merge?

src/wrapper/wrapper.c
```diff
@@ -715,9 +719,13 @@ real_call:; // semicolon here to avoid the error
     }
   #endif // DYAD_SYNC_DIR
 
-    rc = func_ptr (fd);
+    int wronly = is_wronly (fd);
 
-    if (to_sync) {
+    if (wronly == -1)
+        DPRINTF ("Failed to check the mode of the file with fcntl: %s\n", strerror (errno));
```
JaeseungYeom, line 725: There needs to be {} because DPRINTF is macro that may be undefined.

src/wrapper/wrapper.c
```diff
@@ -775,17 +786,23 @@ int fclose (FILE *fp)
...
         sync_directory (path);
     }
   #endif // DYAD_SYNC_DIR
 
-    rc = func_ptr (fp);
+    int wronly = is_wronly (fd);
 
-    if (to_sync) {
+    if (wronly == -1)
+        DPRINTF ("Failed to check the mode of the file with fcntl: %s\n", strerror (errno));
```
JaeseungYeom, line 801: {} needed.

src/wrapper/wrapper.c
ilumsden: Thanks for the catch. Fixed.

src/wrapper/wrapper.c
ilumsden: Thanks for the catch. Fixed.

PR #21, opened 2023-02-22
Build: Fix error with unknown header location

JaeseungYeom:
I had to add the flux header location to build successfully.


ilumsden (approved):
Thanks for catching this. Didn't think about the fact that the API libraries would need access to Flux headers due to the logging macros.

PR #22, opened 2023-02-23
Initialization fixes and minor changes

JaeseungYeom:
- Initialization fixes both in core, wrapper and stream
- Variable name change in the dyad_stream_api: `m_ctx` -> `m_core`
- Function name change in the wrapper. `dyad_sync_start` -> `dyad_wrapper_start`
- Removed unused code about `DYAD_SYNC_START`
- Added `core/dyad_envs.h` that includes all the environment variable names and macro definitions
- Changed the file name `dyad_err.h` to `dyad_rc.h`
- Unify the macro guard format in header files.
- Changes in `Makefile.am` to avoid installing static library version of the module `dyad.a` and teh wrapper.


src/core/Makefile.am
```diff
@@ -1,10 +1,10 @@
...
 libdyad_core_la_LIBADD = $(top_builddir)/src/utils/libutils.la $(top_builddir)/src/utils/libmurmur3.la $(FLUX_CORE_LIBS)
-libdyad_core_la_CPPFLAGS = $(FLUX_CORE_CFLAGS) -I$(top_builddir)/src/utils
+libdyad_core_la_CPPFLAGS = $(AM_CPPFLAGS) -I$(top_builddir)/src/utils $(FLUX_CORE_CFLAGS)
 if PERFFLOW
 libdyad_core_la_LIBADD += $(PERFFLOW_LIBS)
 libdyad_core_la_CPPFLAGS += $(PERFFLOW_PLUGIN_CPPFLAGS) $(PERFFLOW_CFLAGS) -DDYAD_PERFFLOW=1
 endif
 
-include_HEADERS = dyad_core.h dyad_err.h dyad_flux_log.h
+include_HEADERS = dyad_core.h dyad_rc.h dyad_flux_log.h
```
ilumsden, line 10: Need to add `dyad_envs.h` to `include_HEADERS`

src/modules/Makefile.am
```diff
@@ -25,3 +25,6 @@ urpc_la_LIBADD += $(PERFFLOW_LIBS)
 urpc_la_CPPFLAGS += $(PERFFLOW_PLUGIN_CPPFLAGS) $(PERFFLOW_CFLAGS) -DDYAD_PERFFLOW=1
 endif
 endif
+
+install-exec-hook:
+	@(cd $(DESTDIR)$(libdir) && $(RM) $(lib_LTLIBRARIES) dyad.a)
```
ilumsden, line 30: I think it's fine to use `$(lib_LTLIBRARIES)` here, but it might be more clear to use `dyad.la` instead.

src/stream/dyad_stream_api.hpp
```diff
@@ -254,7 +254,7 @@ basic_ifstream_dyad<_CharT, _Traits>& basic_ifstream_dyad<_CharT, _Traits>::
 operator= (basic_ifstream_dyad&& rhs)
 {
     m_stream = std::move (rhs.m_stream);
-    m_ctx = std::move (rhs.ctx);
+    m_core = std::move (rhs.core);
```
ilumsden, line 257: `rhs.core` might be a bug. Shouldn't it be `rhs.m_core`?

src/stream/dyad_stream_api.hpp
```diff
@@ -187,16 +187,16 @@ bool basic_ifstream_dyad<_CharT, _Traits>::is_open ()
 #else  //-----------------------------------------------------------------------
 template <typename _CharT, typename _Traits>
 basic_ifstream_dyad<_CharT, _Traits>::basic_ifstream_dyad (
-    const dyad_stream_core& ctx)
+    const dyad_stream_core& core)
 {
-    m_ctx.init ();
+    m_core.init ();
```
ilumsden, line 192: `m_core.init ();` should be removed. In its place, you should have `: m_core (core)` before the opening curly brace.

src/wrapper/Makefile.am
```diff
@@ -1,13 +1,18 @@
 lib_LTLIBRARIES = libdyad_sync.la
 libdyad_sync_la_SOURCES = wrapper.c
+libdyad_sync_la_LDFLAGS = $(AM_LDFLAGS) -module -avoid-version -no-undefined -shared
```
ilumsden, line 3: Should `-module` be added here? We haven't built `libdyad_sync.so` as a module before.

The other flags all make sense. Just a bit concerned about `-module`.

src/wrapper/wrapper.c
```diff
@@ -249,13 +217,14 @@ real_call:;
...
-        dyad_sync_init ();
-    }
     char *error = NULL;
     typedef FILE *(*fopen_ptr_t) (const char *, const char *);
     fopen_ptr_t func_ptr = NULL;
 
+    if (ctx == NULL) {
+        dyad_wrapper_init ();
+    }
+
```
ilumsden, line 227: If we're going to move this check after the variable initialization for `fopen`, then we should do the same for `open`.

ilumsden: The same should probably be true for `close` and `fclose`. Currently, `fopen` is the only function that does not start with this check.

ilumsden (changes requested):
This looks really good. Just a few relatively minor comments to take a look at.

src/stream/dyad_stream_api.hpp
JaeseungYeom: Correct. We need a test for compilation since this is a header.

src/core/Makefile.am
JaeseungYeom: Added

src/stream/dyad_stream_api.hpp
JaeseungYeom: Good catch.

src/wrapper/Makefile.am
JaeseungYeom: I think that would be better with a name change to `dyad_wrap(per)` instead of `libdyad_sync` because it is not supposed to be linked. What do think?

src/modules/Makefile.am
JaeseungYeom: making that change.

src/wrapper/Makefile.am
```diff
@@ -1,13 +1,18 @@
...
 libdyad_sync_la_LIBADD = $(top_builddir)/src/core/libdyad_core.la
-libdyad_sync_la_CPPFLAGS = -DDYAD_CHECK=1 $(FLUX_CORE_CFLAGS) -I$(top_builddir)/src/utils -I$(top_builddir)/src/core
+libdyad_sync_la_CPPFLAGS = -I$(top_builddir)/src/utils -I$(top_builddir)/src/core $(FLUX_CORE_CFLAGS)
+
+install-exec-hook:
+	@(cd $(DESTDIR)$(libdir) && $(RM) $(lib_LTLIBRARIES))
```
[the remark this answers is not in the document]
JaeseungYeom: changing

src/wrapper/wrapper.c
JaeseungYeom: Good point. I made similar changes to open, close, and fclose.

src/wrapper/Makefile.am
JaeseungYeom: we will address in the followup PR.

ilumsden (approved):
Looks good!

PR #24, opened 2023-02-28
Adds a Data Transport Layer to DYAD to support different ways of transferring data

ilumsden:
This PR adds a new Data Transport Layer (DTL) to the DYAD module and DYAD Core library to allow us to support different ways to transfer data (i.e., transfer backends) between producer and consumer. Through this DTL, this PR also adds a new [UCX](https://openucx.org/)-based transfer backend.

To set the backends, users will set the new `DYAD_DTL_MODE` environment variable. This variable must be set to the same value for both the clients (i.e., user applications that use the C or C++ APIs) and the DYAD module. Valid values are:
* `FLUX_RPC` (~default~): uses Flux's RPC framework to transfer data. This is the same as before
* `UCX` (default): uses UCX to transfer data

The code for the DTL can be found in `src/core/dtl` (for DYAD Core's side) and `src/modules/dtl` (for the DYAD module).

This PR also adds a couple of QOL improvements. Most notably, it adds a new `dyad_init_env` function to the Core library. This function will initialize the DYAD context (`dyad_ctx_t`) using the environment variables defined in `src/core/dyad_env.h`. By providing this function in Core, we can define a baseline environment variable-based initialization for all APIs.


ilumsden:
@JaeseungYeom I've removed the last bit of excess logging and retested. So, this PR is now ready-for-review.

ilumsden:
Thanks to some discussion with @grondo, we identified that a lot of symbols are being exported from DYAD's libraries that shouldn't be exported. I am in the process of correcting this.

One question related to symbol exporting @JaeseungYeom: what do we want to do about `dyad_sync_directory`? That function is conditionally compiled, and, as far as I know, there's no way to tell libtool to conditionally export a symbol.

JaeseungYeom:
I think you can treat dyad_sync_directory similarly to other DAYD internal functions.

ilumsden:
Notes from review:
- [ ] Refactor DTL interfaces to use struct  + factory method (new PR for this)
- [ ] Merge two DTL libraries into one (new PR for this; probably merged with one above)
- [ ] Cache endpoints w/ LRU replacement (new PR for this)
- [ ] Move tag creation to broker, and make tag (broker_rank << 32) | counter (counter is unique ID in broker)
- [ ] Add parens around UCX_CHECK and arg to UCX_CHECK
- [ ] Add timeout to wait for UCX send/recv
- [ ] Add delay of 10 us (i.e., usleep) if `ucp_worker_progress` takes less than 10 us
- [ ] Check how timeouts work in UCX. If I can't find anything, at least add a TODO
- [ ] Send ACK RPC response to validate that RPC correctly reached UCX send

JaeseungYeom:
1. Send ACK RPC response to validate that RPC correctly reached UCX send
2. Add timeout to wait for UCX send/recv
3. Check how timeouts work in UCX. If I can't find anything, at least add a TODO

If 3 is available and 2 is working, do we still need 1?

hariharan-devarajan:
> 1. Send ACK RPC response to validate that RPC correctly reached UCX send
> 2. Add timeout to wait for UCX send/recv
> 3. Check how timeouts work in UCX. If I can't find anything, at least add a TODO
> 
> If 3 is available and 2 is working, do we still need 1?

I believe they give different errors:
1. identifies whether we were able to setup ucx endpoint for now doing data transfer.
2. and 3. to me are ways to know if ucx transfer failed. I think if ucx has timeouts we should use that (that is 3) if not we should build some simple version of that (that is 2)

JaeseungYeom:
> 
I agree with that trying to see if option 3 is available first and then do option 2. For 1, it sounds to me it does not add any other functionality if either 2 or 3 works. If endpoint setup fails, would the entire DYAD shutdown or do we expect to retry? If the latter, I see it could be useful.

JaeseungYeom (commented):
You should add ```#if defined(__cplusplus)
extern "C" {
#endif  // defined(__cplusplus)

#if defined(__cplusplus)
};
#endif  // defined(__cplusplus)``` to headers. I am still going through this PR, but you can start addressing.

src/dtl/dyad_dtl_impl.c
```diff
@@ -0,0 +1,82 @@
...
+        return DYAD_RC_OK;
+    }
+    return DYAD_RC_BADDTLMODE;
+}
+
+dyad_rc_t dyad_dtl_finalize(dyad_dtl_t **dtl_handle)
+{
+    dyad_rc_t rc = DYAD_RC_OK;
+    if (dtl_handle == NULL || *dtl_handle == NULL)
+        return DYAD_RC_OK;
```
JaeseungYeom, line 61: Add an inline comment why this case is considered ok.

src/dtl/flux_dtl.c
```diff
@@ -0,0 +1,144 @@
...
+{
+    int errcode = 0;
+    errcode = flux_request_unpack (
+        msg,
+        NULL,
+        "{s:s}",
+        "upath",
+        upath
+    );
+    if (errcode < 0) {
```
JaeseungYeom, line 45: TODO: Check this with a macro as it is done some other places. Also, rc is better than errorcode.

JaeseungYeom: There are similar lines.

src/dtl/dyad_dtl.h
```diff
@@ -0,0 +1,20 @@
+#ifndef DYAD_DTL_H
+#define DYAD_DTL_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+    
+enum dyad_dtl_mode {
+    DYAD_DTL_UCX = 0,
+    DYAD_DTL_FLUX_RPC = 1,
```
JaeseungYeom, line 10: add something like DYAD_DTL_END without comma at the end

src/modules/dyad.c
```diff
@@ -126,15 +127,41 @@ dyad_fetch_request_cb (flux_t *h,
     char *upath = NULL;
     char fullpath[PATH_MAX + 1] = {'\0'};
     int saved_errno = errno;
-    errno = 0;
+    dyad_rc_t rc = 0;
 
-    if (flux_msg_get_userid (msg, &userid) < 0)
-        goto error;
+    if (!flux_msg_is_streaming (msg)) {
```
JaeseungYeom, line 132: We should have a discussion on what the best terms is for "streaming" later.

src/dtl/dyad_dtl_impl.c
```diff
@@ -0,0 +1,82 @@
...
+            return rc;
+        }
+        (*dtl_handle)->rpc_pack = dyad_dtl_ucx_rpc_pack;
+        (*dtl_handle)->rpc_unpack = dyad_dtl_ucx_rpc_unpack;
+        (*dtl_handle)->rpc_respond = dyad_dtl_ucx_rpc_respond;
+        (*dtl_handle)->rpc_recv_response = dyad_dtl_ucx_rpc_recv_response;
+        (*dtl_handle)->establish_connection = dyad_dtl_ucx_establish_connection;
+        (*dtl_handle)->send = dyad_dtl_ucx_send;
+        (*dtl_handle)->recv = dyad_dtl_ucx_recv;
+        (*dtl_handle)->close_connection = dyad_dtl_ucx_close_connection;
```
JaeseungYeom, line 32: These should move into dtad_dtl_ucx_init()

src/dtl/dyad_dtl_impl.c
```diff
@@ -0,0 +1,82 @@
...
+            return rc;
+        }
+        (*dtl_handle)->rpc_pack = dyad_dtl_flux_rpc_pack;
+        (*dtl_handle)->rpc_unpack = dyad_dtl_flux_rpc_unpack;
+        (*dtl_handle)->rpc_respond = dyad_dtl_flux_rpc_respond;
+        (*dtl_handle)->rpc_recv_response = dyad_dtl_flux_rpc_recv_response;
+        (*dtl_handle)->establish_connection = dyad_dtl_flux_establish_connection;
+        (*dtl_handle)->send = dyad_dtl_flux_send;
+        (*dtl_handle)->recv = dyad_dtl_flux_recv;
+        (*dtl_handle)->close_connection = dyad_dtl_flux_close_connection;
```
JaeseungYeom, line 51: These should move into dyad_dtl_flux_init()

JaeseungYeom:
Don't forget to run clang formatter

configure.ac
```diff
@@ -153,7 +185,7 @@ AC_CHECK_FUNCS( \
 # Add any necessary compilation flags #
 #######################################
 if test "x$enable_debug" = xyes; then
-    CPPFLAGS="$CPPFLAGS -DDYAD_FULL_DEBUG=1 -DDYAD_LOGGING_ON=1"
+    CFLAGS="$CFLAGS -DDYAD_FULL_DEBUG=1 -DDYAD_LOGGING_ON=1"
```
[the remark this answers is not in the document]
ilumsden: Yes. The use of `CPPFLAGS` was the main reason why the symbol visibility stuff from Tuesday wasn't working. By being more specific with `CFLAGS` and `CXXFLAGS`, things will actually work as expected.

This does remind me that I need to add a similar line for `CXXFLAGS` here though. I'll do that.

src/core/dyad_core.c
```diff
@@ -565,6 +649,96 @@ dyad_rc_t dyad_init (bool debug,
...
+        if (strncmp (e, "FLUX_RPC", dtl_mode_env_len) == 0) {
+            dtl_mode = DYAD_DTL_FLUX_RPC;
+        } else if (strncmp (e, "UCX", dtl_mode_env_len) == 0) {
+            dtl_mode = DYAD_DTL_UCX;
+        } else {
+            if (debug) {
+                fprintf (stderr, "Invalid DTL mode provided through %s. \
+                        Defaulting to UCX\n", DYAD_DTL_MODE_ENV);
+            }
+            dtl_mode = DYAD_DTL_UCX;
```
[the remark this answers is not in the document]
ilumsden: I think having UCX as default makes sense for the time being. Most systems nowadays have UCX, mainly because its used by MPI. Also, right now, the UCX dependency is required, so DYAD won't build without UCX.

We can think more about this later, but I think that should be a separate PR if we change it.

src/core/dyad_core.c
```diff
@@ -310,73 +316,133 @@ fetch_done:;
...
-                        "upath", kvs_data->fpath);
-    // If the RPC dispatch failed, log an error and return DYAD_BADRPC
-    if (*f == NULL) {
-        DYAD_LOG_ERR (ctx, "Cannot send RPC to producer plugin!\n");
-        return DYAD_RC_BADRPC;
+    dyad_rc_t final_rc = DYAD_RC_OK;
+    flux_future_t *f;
+    json_t* rpc_payload;
+    DYAD_LOG_INFO (ctx, "Packing payload for RPC to DYAD module");
+    rc = ctx->dtl_handle->rpc_pack (
```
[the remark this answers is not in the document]
ilumsden: I agree its awkward. I spent quite a bit of time looking for an alternative, but I couldn't find one.

A macro could work for simplifying this, but I'd highly recommend against a global variable. We don't necessarily want the context to always be shared across functions.

src/core/dyad_core.h
```diff
@@ -32,21 +33,22 @@ extern "C" {
...
-    char* prod_managed_path;  // producer path managed by DYAD
-    char* cons_managed_path;  // consumer path managed by DYAD
+    flux_t* h;                     // the Flux handle for DYAD
+    struct dyad_dtl *dtl_handle;   // Opaque handle to DTL info
+    bool debug;                    // if true, perform debug logging
+    bool check;                    // if true, perform some check logging
+    bool reenter;                  // if false, do not recursively enter DYAD
+    bool initialized;              // if true, DYAD is initialized
+    bool shared_storage;           // if true, the managed path is shared
+    bool sync_started;             // TODO
```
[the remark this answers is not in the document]
ilumsden: I'll see if it's used anywhere. Regardless, I think it makes sense to remove it. Synchronizing producer and consumer application starts isn't really DYAD's job, in my opinion

src/core/dyad_core.h
```diff
@@ -32,21 +33,22 @@ extern "C" {
...
+    uint32_t rank;                 // Flux rank for DYAD
+    char* kvs_namespace;           // Flux KVS namespace for DYAD
+    char* prod_managed_path;       // producer path managed by DYAD
+    char* cons_managed_path;       // consumer path managed by DYAD
 };
-extern const struct dyad_ctx dyad_ctx_default;
+DYAD_DLL_EXPORTED extern const struct dyad_ctx dyad_ctx_default;
 typedef struct dyad_ctx dyad_ctx_t;
 
 struct dyad_kvs_response {
```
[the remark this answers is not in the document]
ilumsden: Good point. This is a holdover from an earlier design of core that slipped through the cracks. I'll move it.

src/core/dyad_core.c
```diff
@@ -310,73 +316,133 @@ fetch_done:;
...
+    size_t* file_len)
 #else
-static inline dyad_rc_t dyad_rpc_get (
+static inline dyad_rc_t dyad_get_data (
     const dyad_ctx_t* ctx,
     const dyad_kvs_response_t* restrict kvs_data,
     const char** file_data,
-    int* file_len,
-    flux_future_t** f)
+    size_t* file_len)
```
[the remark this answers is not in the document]
JaeseungYeom: This can be a separate PR if that works better

src/dtl/flux_dtl.c
ilumsden: This check should **not** use `DYAD_IS_ERROR` because it's not checking a DYAD return code. It's checking a Flux return code. The choice of variable name was also intentional to emphasize this difference.

src/dtl/dyad_dtl_impl.h
```diff
@@ -0,0 +1,64 @@
...
+union dyad_dtl_private {
+    struct dyad_dtl_ucx* ucx_dtl_handle;
+    struct dyad_dtl_flux* flux_dtl_handle;
+};
+typedef union dyad_dtl_private dyad_dtl_private_t;
+
+enum dyad_dtl_comm_mode {
+    DYAD_COMM_NONE = 0, // Sanity check value for when connection isn't established
+    DYAD_COMM_RECV = 1, // DTL connection will only receive data
+    DYAD_COMM_SEND = 2, // DTL connection will only send data
```
JaeseungYeom, line 34: add DYAD_COMM_END = 3 and no comma at the end

src/dtl/dyad_dtl.h
ilumsden: Why? I'm not sure why we'd need a `DYAD_DTL_END`. The `dyad_dtl_mode` enum is used to select which underlying implementation of the DTL we're going to use.

src/modules/dyad.c
ilumsden: I agree. FWIW, in this context, the term "streaming" refers to Flux's streaming RPCs (i.e., an RPC which can send multiple return messages and is ended with an error return message with errno set to `ENODATA`).

src/dtl/dyad_dtl_impl.c
ilumsden: Actually, I'd argue they should stay here. The purpose of `dyad_dtl_ucx_init` (and its Flux RPC equivalent) is to do any setup and initialization of the underlying handle (e.g., `dyad_dtl_ucx_t`). On the other hand, the purpose of the top level initializer (`dyad_dtl_init`) is to initialize the "top-level" handle (i.e., `dyad_dtl_t`).

configure.ac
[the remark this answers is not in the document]
ilumsden: Latest commits resolved this

src/core/dyad_core.c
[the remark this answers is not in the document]
ilumsden: I added restrict to the latest commits. However, I should mention that, due to stuff with the Flux handle, I can't add it in many places without telling the compiler incorrect things (which will break the build). I think there were only 3 places where I could definitely add it.

src/core/dyad_core.h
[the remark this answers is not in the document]
ilumsden: Most recent commits have removed this

src/core/dyad_core.h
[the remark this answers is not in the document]
ilumsden: Moved this in `dyad_core.c` with the last two commits

src/dtl/dyad_dtl_impl.c
ilumsden: Added

src/dtl/dyad_dtl.h
ilumsden: Thanks for clarifying over Slack. I've added this

src/dtl/dyad_dtl_impl.c
ilumsden: Discussed over Slack and added

src/dtl/dyad_dtl_impl.h
ilumsden: Discussed over Slack and added

src/core/dyad_core.c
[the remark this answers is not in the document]
JaeseungYeom: libfabric is another to consider. AWS EFA is based on libfabric for example. So, we should definitely detect what is available and use it as the default instead of assuming UCX availability.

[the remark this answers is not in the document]
JaeseungYeom: This is can be a followup PR. Even if libfabric is not supported, UCX availability should be checked.

src/core/dyad_core.c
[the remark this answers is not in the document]
JaeseungYeom: I do not like global variable approach as well. However, context does not need to be shared for the sake of that. Only dtl handle need to be global, which will be read-only once initialized.

src/core/dyad_core.c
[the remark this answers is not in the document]
JaeseungYeom: You can show me the problem later.

src/dtl/flux_dtl.c
JaeseungYeom: It does not need to be DYAD_IS_ERROR. It can be FLUX_IS_ERROR for example. error code is misleading because it is not necessarily a code for error like errorno.

src/core/dyad_core.c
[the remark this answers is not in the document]
JaeseungYeom: Requiring UCX as the dependency is also problematic in the short term. Vanessa would not be able to try this if that is the case.

src/dtl/flux_dtl.c
ilumsden: Just pushed a new commit that deals with this.

src/core/dyad_core.c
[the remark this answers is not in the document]
ilumsden: Vanessa would just have to install UCX. UCX can work on cloud (although I can't say anything about performance). If a HPC-grade network (e.g., InfiniBand) is not present, UCX will just fall back to TCP/IP.

src/core/dyad_core.c
```diff
@@ -30,15 +31,28 @@ const struct dyad_ctx dyad_ctx_default = {
...
 static int gen_path_key (const char* str,
                          char* path_key,
                          const size_t len,
                          const uint32_t depth,
                          const uint32_t width)
 {
-    static const uint32_t seeds[10] = {104677u, 104681u, 104683u, 104693u,
-                                       104701u, 104707u, 104711u, 104717u,
-                                       104723u, 104729u};
+    static const uint32_t seeds[10] = {104677u,
```
JaeseungYeom, line 46: These individual values are not that meaningful, and you almost need not seem them. No need to make this change. Is this from clang format?

src/core/dyad_core.c
```diff
@@ -69,12 +83,11 @@ static int gen_path_key (const char* str,
 }
 
 #if DYAD_PERFFLOW
-__attribute__ ((annotate ("@critical_path()")))
-static dyad_rc_t dyad_kvs_commit (const dyad_ctx_t* ctx,
-                                  flux_kvs_txn_t* txn)
+__attribute__ ((annotate ("@critical_path()"))) static dyad_rc_t dyad_kvs_commit (
```
JaeseungYeom, line 86: This is actually where a new line is needed.

src/core/dyad_core.c
```diff
@@ -518,8 +586,23 @@ dyad_rc_t dyad_init (bool debug,
...
+    // Initialize the DTL based on the value of dtl_mode
+    // If an error occurs, log it and return an error
+    FLUX_LOG_INFO ((*ctx)->h, "DYAD_CORE: inintializing DYAD DTL");
+    rc = dyad_dtl_init(
+        &(*ctx)->dtl_handle,
+        dtl_mode,
+        (*ctx)->h,
+        (*ctx)->debug
+    );
+    if (DYAD_IS_ERROR(rc))
```
JaeseungYeom: space

src/core/dyad_core.c
```diff
@@ -420,10 +486,8 @@ static inline dyad_rc_t dyad_pull (const dyad_ctx_t* restrict ctx,
     rc = DYAD_RC_OK;
 
 pull_done:
-    // Destroy the Flux future for the RPC, if needed
-    if (f != NULL) {
-        flux_future_destroy (f);
-        f = NULL;
+    if (file_data != NULL) {
+        free(file_data);
```
JaeseungYeom: space

src/core/dyad_core.c
```diff
@@ -310,73 +316,133 @@ fetch_done:;
...
     memset (file_path, 0, PATH_MAX + 1);
     memset (file_path_copy, 0, PATH_MAX + 1);
 
-    // Call dyad_rpc_get to dispatch a RPC to the producer's Flux broker
+    // Call dyad_get_data to dispatch a RPC to the producer's Flux broker
     // and retrieve the data associated with the file
-    rc = dyad_rpc_get (ctx, kvs_data, &file_data, &file_len, &f);
-    if (rc != DYAD_RC_OK) {
+    rc = dyad_get_data (ctx, kvs_data, &file_data, &file_len);
+    if (DYAD_IS_ERROR(rc)) {
```
JaeseungYeom: space

src/core/dyad_core.c
```diff
@@ -310,73 +316,133 @@ fetch_done:;
...
+    {
+        DYAD_LOG_ERR(ctx, "Cannot receive and/or parse the RPC response\n");
+        goto get_done;
+    }
+    DYAD_LOG_INFO (ctx, "Establish DTL connection with DYAD module");
+    rc = ctx->dtl_handle->establish_connection (
+        ctx->dtl_handle,
+        DYAD_COMM_RECV
+    );
+    if (DYAD_IS_ERROR(rc)) {
```
JaeseungYeom: space

src/core/dyad_core.c
```diff
@@ -310,73 +316,133 @@ fetch_done:;
...
+        goto get_done;
+    }
+    DYAD_LOG_INFO (ctx, "Receive RPC response from DYAD module");
+    rc = ctx->dtl_handle->rpc_recv_response (
+        ctx->dtl_handle,
+        f
+    );
+    if (DYAD_IS_ERROR(rc))
+    {
+        DYAD_LOG_ERR(ctx, "Cannot receive and/or parse the RPC response\n");
```
JaeseungYeom: space

src/core/dyad_core.c
```diff
@@ -310,73 +316,133 @@ fetch_done:;
...
+    DYAD_LOG_INFO (ctx, "Packing payload for RPC to DYAD module");
+    rc = ctx->dtl_handle->rpc_pack (
+        ctx->dtl_handle,
+        kvs_data->fpath,
+        kvs_data->owner_rank,
+        &rpc_payload
+    );
+    if (DYAD_IS_ERROR(rc))
+    {
+        DYAD_LOG_ERR(ctx, "Cannot create JSON payload for Flux RPC to DYAD module\n");
```
JaeseungYeom: space

src/core/dyad_core.c
```diff
@@ -296,6 +295,13 @@ static inline dyad_rc_t dyad_fetch (const dyad_ctx_t* restrict ctx,
         goto fetch_done;
     }
     (*resp)->fpath = malloc (strlen (upath) + 1);
+    if ((*resp)->fpath == NULL)
+    {
+        DYAD_LOG_ERR (ctx, "Cannot allocate a buffer for the file path in the dyad_kvs_response_t object\n");
+        free(*resp);
```
JaeseungYeom: free (

src/dtl/dyad_flux_log.h
```diff
@@ -17,11 +21,14 @@
     do {                            \
     } while (0)
 #else
-#define DYAD_LOG_INFO(dyad_ctx, ...) \
-    flux_log (dyad_ctx->h, LOG_INFO, __VA_ARGS__)
+#define DYAD_LOG_INFO(dyad_ctx, ...) flux_log (dyad_ctx->h, LOG_INFO, __VA_ARGS__)
```
JaeseungYeom, line 24: space

src/dtl/flux_dtl.c
```diff
@@ -0,0 +1,140 @@
...
+
+dyad_rc_t dyad_dtl_flux_establish_connection (dyad_dtl_t* self,
+                                              dyad_dtl_comm_mode_t comm_mode)
+{
+    return DYAD_RC_OK;
+}
+
+dyad_rc_t dyad_dtl_flux_send (dyad_dtl_t* self, void* buf, size_t buflen)
+{
+    int errcode = 0;
```
JaeseungYeom, line 79: variable naming

src/dtl/flux_dtl.c
```diff
@@ -0,0 +1,140 @@
...
+    FLUX_LOG_INFO (self->private.flux_dtl_handle->h,
+                   "Send data to consumer using a Flux RPC response");
+    errcode = flux_respond_raw (self->private.flux_dtl_handle->h,
+                                self->private.flux_dtl_handle->msg,
+                                buf,
+                                (int)buflen);
+    if (errcode < 0) {
```
JaeseungYeom, line 86: need macro

src/dtl/flux_dtl.c
```diff
@@ -0,0 +1,140 @@
...
+    dyad_dtl_flux_t* dtl_handle = self->private.flux_dtl_handle;
+    FLUX_LOG_INFO (dtl_handle->h, "Get file contents from module using Flux RPC\n");
+    if (dtl_handle->f == NULL) {
+        FLUX_LOG_ERR (dtl_handle->h,
+                      "Cannot get data using RPC without a Flux future\n");
+        // TODO create new RC for this
+        return DYAD_RC_FLUXFAIL;
+    }
+    rc = flux_rpc_get_raw (dtl_handle->f, (const void**)buf, (int*)buflen);
+    if (rc < 0) {
```
JaeseungYeom, line 112: need macro

src/dtl/dyad_rc.h
```diff
@@ -0,0 +1,50 @@
...
+                                   // end of stream) sooner than expected
+    DYAD_RC_BAD_B64DECODE = -18,   // Decoding of data w/ base64 failed
+    DYAD_RC_BAD_COMM_MODE = -19,   // Invalid comm mode provided to DTL
+};
+
+typedef enum dyad_core_return_codes dyad_rc_t;
+
+#define DYAD_IS_ERROR(code) ((code) < 0)
+
+#define FLUX_IS_ERROR(code) ((code) < 0)
```
JaeseungYeom, line 44: space

src/utils/base64/base64.c
```diff
@@ -0,0 +1,253 @@
...
+	return dest_offset;
+}
+
+size_t base64_decoded_length(size_t srclen)
+{
+	return ((srclen+3)/4*3);
+}
+
+ssize_t base64_decode_quartet_using_maps(const base64_maps_t *maps, char dest[3],
+				     const char src[4])
```
JaeseungYeom, line 122: align

src/utils/base64/base64.c
```diff
@@ -0,0 +1,253 @@
...
+	dest[0] = (a << 2) | (b >> 4);
+	dest[1] = ((b & 0xf) << 4) | (c >> 2);
+	dest[2] = ((c & 0x3) << 6) | d;
+
+	return 0;
+}
+
+
+ssize_t base64_decode_tail_using_maps(const base64_maps_t *maps, char dest[3],
+				  const char * src, const size_t srclen)
```
JaeseungYeom, line 147: align

src/core/dyad_core.c
ilumsden: Yes

src/core/dyad_core.c
ilumsden: This is a clang format thing. I'll fix it.

src/dtl/dyad_flux_log.h
ilumsden: Since there are not spaces after function-like macros in `flux-core`, I'd argue that we shouldn't add spaces in DYAD either.

src/utils/utils.c
```diff
@@ -160,11 +159,9 @@ bool cmp_prefix (const char* __restrict__ prefix,
...
-            ((prefix + strlen (prefix) <= (char*)u_len)
-             || (u_len_end <= prefix))
+            ((prefix + strlen (prefix) <= (char*)u_len) || (u_len_end <= prefix))
             && ((full + strlen (full) <= (char*)u_len) || (u_len_end <= full))
-            && ((delim + strlen (delim) <= (char*)u_len)
-                || (u_len_end <= delim));
+            && ((delim + strlen (delim) <= (char*)u_len) || (u_len_end <= delim));
 
         if (!no_overlap) {
             DPRINTF ("DYAD UTIL: buffers overlap.\n");
```
hariharan-devarajan, line 167: Why are these logging different than the login we do in DYAD like DYAD_LOG_INFO

src/utils/utils.c
```diff
@@ -232,9 +229,8 @@ bool cmp_canonical_path_prefix (const char* __restrict__ prefix,
...
     {
         const char* const upath_end = upath + upath_capacity;
-        bool no_overlap =
-            ((prefix + strlen (prefix) <= upath) || (upath_end <= prefix))
-            && ((path + strlen (path) <= upath) || (upath_end <= path));
+        bool no_overlap = ((prefix + strlen (prefix) <= upath) || (upath_end <= prefix))
+                          && ((path + strlen (path) <= upath) || (upath_end <= path));
 
         if (!no_overlap) {
             DPRINTF ("DYAD UTIL: buffers overlap\n");
```
hariharan-devarajan, line 236: Again the logging is not using DYAD_LOG_INFO

src/utils/utils.c
```diff
@@ -339,13 +337,14 @@ int mkdir_as_needed (const char* path, const mode_t m)
...
+                    (sb.st_mode & RWX_UGO),
+                    (m & RWX_UGO));
                 return 5;  // already exists but with different mode
             }
             return 1;  // already exists
         }
-        DPRINTF ("Cannot create directory \"%s\": %s\n", path,
-                 strerror (errno));
+        DPRINTF ("Cannot create directory \"%s\": %s\n", path, strerror (errno));
         perror ("mkdir_as_needed() ");
```
hariharan-devarajan, line 348: Again we should have DYAD_LOG_ERROR defined.

src/wrapper/Makefile.am
```diff
@@ -1,17 +1,23 @@
...
 dyad_wrapper_la_SOURCES = wrapper.c
 dyad_wrapper_la_LDFLAGS = \
+	-Wl,-rpath,'$(UCX_LIBDIR)' \
 	$(AM_LDFLAGS) \
 	-module \
 	-avoid-version \
 	-no-undefined \
 	-shared \
-	-export-symbols wrapper.sym
+	# -export-symbols wrapper.sym
```
hariharan-devarajan, line 10: comments have problems in Makefile.am i would recommend removing this line is not needed. Make sure you remove the \ in previous line as wel.

src/wrapper/wrapper.c
```diff
@@ -169,16 +112,16 @@ void dyad_wrapper_fini ()
...
     typedef int (*open_ptr_t) (const char *, int, mode_t, ...);
     open_ptr_t func_ptr = NULL;
     int mode = 0;
 
-    if (ctx == NULL) {
-        dyad_wrapper_init ();
-    }
+    // if (ctx == NULL) {
+    //     dyad_wrapper_init ();
+    // }
```
hariharan-devarajan, line 124: I guess this logic is commented as the init happens on the library load. If so remove the commented lines.

src/wrapper/wrapper.c
```diff
@@ -199,8 +142,7 @@ int open (const char *path, int oflag, ...)
     }
 
     if (!(ctx && ctx->h) || (ctx && !ctx->reenter)) {
-        IPRINTF (ctx, "DYAD_SYNC: open sync not applicable for \"%s\".\n",
-                 path);
+        IPRINTF (ctx, "DYAD_SYNC: open sync not applicable for \"%s\".\n", path);
```
hariharan-devarajan, line 145: We should use the same Logging mechanism DYAD_LOG_INFO

src/wrapper/wrapper.c
```diff
@@ -216,15 +158,15 @@ real_call:;
...
     char *error = NULL;
     typedef FILE *(*fopen_ptr_t) (const char *, const char *);
     fopen_ptr_t func_ptr = NULL;
 
-    if (ctx == NULL) {
-        dyad_wrapper_init ();
-    }
+    // if (ctx == NULL) {
+    //     dyad_wrapper_init ();
+    // }
```
hariharan-devarajan, line 169: Same comment as for open. Probably remove this line.

src/wrapper/wrapper.c
```diff
@@ -238,7 +180,8 @@ FILE *fopen (const char *path, const char *mode)
     }
 
     if (!(ctx && ctx->h) || (ctx && !ctx->reenter) || !path) {
-        IPRINTF (ctx, "DYAD_SYNC: fopen sync not applicable for \"%s\".\n",
+        IPRINTF (ctx,
```
hariharan-devarajan, line 183: DYAD_LOG_INFO

src/wrapper/wrapper.c
```diff
@@ -216,15 +158,15 @@ real_call:;
...
     fopen_ptr_t func_ptr = NULL;
 
-    if (ctx == NULL) {
-        dyad_wrapper_init ();
-    }
+    // if (ctx == NULL) {
+    //     dyad_wrapper_init ();
+    // }
 
     func_ptr = (fopen_ptr_t)dlsym (RTLD_NEXT, "fopen");
```
hariharan-devarajan, line 171: All this logic of moving to next symbol can be abstracted using macro for each case. 
```
#define MAP_OR_FAIL(func_)                                         \
  if (!(real_##func_##_)) {                                        \
    real_##func_##_ = (real_t_##func_##_)dlsym(RTLD_NEXT, #func_); \
    if (!(real_##func_##_)) {                                      \
      fprintf(stderr, "failed to map symbol\n");    \
    }                                                              \
  }
#else
#define MAP_OR_FAIL(func)
#endif
```

I recommend using GOTCHA tool which handles this more dynamically.

hariharan-devarajan (commented):
Here are some of my feedback in the code.

.clang-format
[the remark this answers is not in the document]
ilumsden: There actually are a handful of actual changes to `.clang-format`. Those changes resolve several complications/annoyances, particularly regarding the PerfFlow Aspect annotations messing up return type formatting.

src/core/Makefile.am
```diff
@@ -1,11 +1,29 @@
...
+libdyad_core_la_SOURCES = \
+	dyad_core.c
+libdyad_core_la_LIBADD = \
+	$(top_builddir)/src/dtl/libdyad_dtl.la \
+	$(UCX_LIBS) \
+	$(JANSSON_LIBS) \
+	$(FLUX_CORE_LIBS)
+libdyad_core_la_CFLAGS = \
+	$(AM_CFLAGS) \
+	-I$(top_srcdir)/src/utils \
```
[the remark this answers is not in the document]
ilumsden: I absolutely agree with you @hariharan-devarajan, but I did this because we wanted to stay (relatively) consistent with Flux. This `flux-core` repo does something similar to this in its `Makefile.am`s.

However, regardless of what Flux does and how much we want to follow that, I also believe that any changes we might want to make regarding organization of files should belong in a separate PR.

src/core/dyad_core.c
```diff
@@ -1,8 +1,9 @@
+#include "dyad_core.h"
+
 #include <libgen.h>
 #include <unistd.h>
 
-#include "dyad_core.h"
-#include "dyad_flux_log.h"
+#include "dyad_dtl_impl.h"
```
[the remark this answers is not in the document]
ilumsden: I think I better understand what you're saying about the headers now that I read this comment, but, again, I believe such a change should be its own PR.

src/core/dyad_core.c
```diff
@@ -308,75 +289,112 @@ fetch_done:;
...
+                                    &rpc_payload);
+    if (DYAD_IS_ERROR (rc)) {
+        DYAD_LOG_ERR (ctx,
+                      "Cannot create JSON payload for Flux RPC to DYAD "
+                      "module\n");
+        goto get_done;
+    }
+    DYAD_LOG_INFO (ctx, "Sending payload for RPC to DYAD module");
+    f = flux_rpc_pack (ctx->h,
+                       "dyad.fetch",
```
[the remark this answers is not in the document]
ilumsden: Good idea. I'll do that

src/dtl/Makefile.am
```diff
@@ -0,0 +1,25 @@
...
+    ucx_dtl.h
+libdyad_dtl_la_LIBADD = \
+    $(top_builddir)/src/utils/libutils.la \
+    $(top_builddir)/src/utils/libmurmur3.la \
+    $(UCX_LIBS) \
+    $(JANSSON_LIBS) \
+    $(FLUX_CORE_LIBS)
+libdyad_dtl_la_CFLAGS = \
+    $(AM_CFLAGS) \
+    -I$(top_srcdir)/src/utils \
```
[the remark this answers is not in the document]
ilumsden: Agreed, but, again, this should probably be its own PR

src/utils/utils.c
ilumsden: Not sure. That's been there since I started on DYAD.

src/wrapper/Makefile.am
ilumsden: Thanks for catching this. I'll fix it.

src/wrapper/wrapper.c
ilumsden: Thanks for catching this. I meant to remove it, but it must have slipped under my radar. I'll fix that.

src/wrapper/wrapper.c
ilumsden: I agree that switching the GOTCHA in the future is the right move. That definitely doesn't belong in this PR though. If we don't have one yet, I'll create an issue for switching from "raw" `LD_PRELOAD` to GOTCHA.

src/wrapper/wrapper.c
ilumsden: We should explore unifying the logging mechanisms in DYAD. I'll create an issue for that so we can track it as future work.

src/wrapper/wrapper.c
ilumsden: Issue #43 has been created to track adding GOTCHA support

src/wrapper/wrapper.c
ilumsden: I created #44 to track this.

src/core/dyad_core.c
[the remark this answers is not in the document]
ilumsden: Done

src/wrapper/Makefile.am
ilumsden: Done

src/wrapper/wrapper.c
ilumsden: Done

src/wrapper/wrapper.c
ilumsden: Done

src/utils/read_all.h
```diff
@@ -8,8 +8,8 @@
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
 
-#ifndef DYAD_MODULES_READ_ALL_H
-#define DYAD_MODULES_READ_ALL_H
+#ifndef UTILS_READ_ALL_H
```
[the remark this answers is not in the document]
ilumsden: Done

src/core/dyad_core.c
```diff
@@ -310,73 +319,127 @@ fetch_done:;
...
+        rpc_payload
+    );
+    if (f == NULL)
+    {
+        DYAD_LOG_ERR(ctx, "Cannot send RPC to producer module\n");
+        rc = DYAD_RC_BADRPC;
+        goto get_done;
+    }
+    DYAD_LOG_INFO (ctx, "Receive RPC response from DYAD module");
+    rc = dyad_dtl_recv_rpc_response(ctx->dtl_handle, f);
```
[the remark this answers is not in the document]
ilumsden: We should do this alongside #30

src/wrapper/wrapper.c
JaeseungYeom: It make sense to address it in a separate PR. @hariharan-devarajan please find time with us to explain the benefits from the users perspective as well as developer's.

src/core/dyad_core.c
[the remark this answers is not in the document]
JaeseungYeom: I agree with both of you. It needs to be changed and should be done in a separate PR.

src/core/Makefile.am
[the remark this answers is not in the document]
JaeseungYeom: I agree that we should put more effort to clean up installation and make distinction between public headers and private headers. This can be worked on in a followup PR.

.clang-format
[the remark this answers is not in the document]
JaeseungYeom: I think Ian's approach is reasonable to avoid manual formatting and productivity decrease although I do not like the solution for the long term. We want the annotation line separate from the function signature, and we want to avoid manual formatting. So far, there is no solution known to us. So, this is an alternative.

src/core/dyad_core.c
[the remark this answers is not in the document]
JaeseungYeom: I agree with both of you.

PR #25, opened 2023-03-15
Adds GitHub Action that tests if DYAD compiles

ilumsden:
This PR adds a GitHub Action to test if DYAD compiles. This action will run on the following triggers:
* New commits to the `main` branch
* New commits to a PR that will be merged into the `main` branch

To do this testing, this action will first fetch a flux-core image from DockerHub. Currently, the following images will be used (one per job):
* `latest`
* `focal`
* `focal-v0.48.0`
* `focal-v0.47.0`
* `focal-v0.46.1`

I was thinking we could always test against the `latest` image and the newest Ubuntu image (GitHub requires Docker-based actions to use Ubuntu). Then, we could also test against the Ubuntu-based images for the three most recent versions of Flux.

After pulling the Docker image, all the action does is run the usual commands to build DYAD. If any step fails, the entire run will fail.


JaeseungYeom:
Does the Ubutu-based image have clang, and what version of gcc does it have?

JaeseungYeom (approved):
This is good for now. It will help the development. Eventually, we want to have a setup that test against various compilers.

PR #26, opened 2023-06-22
Hotfix: updates build system to prevent excessive symbol exports

ilumsden:
DYAD's various libraries were previously exporting a lot of symbols that were not supposed to be exported. This PR fixes this using libtool's `-export-symbols` and `-export-symbols-regex` flags.


JaeseungYeom (approved):
Let me know when you add the missing characters (newlines?) at the end of `dyad_core.sym` and `dyad_wrapper.sym`. Otherwise, it looks good. I will merge as soon as those are fixed.

PR #28, opened 2023-07-07
Hotfix: removes blank lines from end of .sym files because they break builds

ilumsden:
During the review of #26, we noticed that GitHub didn't like the lack of a newline at the end of our `.sym` files, so we added those newlines. However, the addition of those newlines breaks Autotools builds. This PR adds those newlines back to the `.sym` files to fix DYAD's build process


JaeseungYeom (approved):
LGTM

PR #38, opened 2023-07-27
Gets the compile test action working

### LINKED ISSUE #35: Fix Docker pulling for CI
I recently realized that the CI for ensuring DYAD builds was pulled into `main` a bit too early. It is currently failing in all cases due to being unable to pull the specified Docker image from DockerHub.

I'll look into this (and other possible CI issues) and fix it.

ilumsden:
Closes #35 

The GitHub Action for testing the compilation works was merged a bit too early. This PR fixes the issues that were cropping up with the action.

### DIFF
--- a/.github/workflows/compile_test.yaml
+++ b/.github/workflows/compile_test.yaml
@@ -32,7 +32,7 @@ jobs:
         run: ./autogen.sh
 
       - name: Configure build
-        run: ./configure --enable-dyad-debug CC=$(which gcc) CXX=$(which g++)
+        run: ./configure --enable-dyad-debug
 
       - name: Build DYAD
         run: make -j
--- a/.github/workflows/compile_test.yaml
+++ b/.github/workflows/compile_test.yaml
@@ -10,14 +10,21 @@ jobs:
   compile:
     strategy:
       matrix:
-        flux-images:
-          - latest
+        base-os:
           - focal
-          - focal-v0.48.0
-          - focal-v0.47.0
-          - focal-v0.46.1
+          - fedora38
+          - fedora35
+          - el8
+          - el7
+        flux-version:
+          - v0.52.0
+          - v0.51.0
+          - v0.50.0
+          - v0.49.0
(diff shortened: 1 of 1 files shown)


.github/workflows/compile_test.yaml
```diff
@@ -10,22 +10,29 @@ jobs:
...
+    container: fluxrm/flux-core:${{ matrix.base-os }}-${{ matrix.flux-version }}
     steps:
       - uses: actions/checkout@v2
 
       - name: Run autogen
         run: ./autogen.sh
 
       - name: Configure build
-        run: ./configure --enable-dyad-debug CC=gcc CXX=g++
+        run: ./configure --enable-dyad-debug CC=$(which gcc) CXX=$(which g++)
```
JaeseungYeom, line 35: It would be better if you just rely on env variables.

JaeseungYeom (approved):

.github/workflows/compile_test.yaml
ilumsden: From what I understand, putting things like `CC` and `CXX` after `configure` is usually preferred because it is more portable. If you set `CC` and `CXX` through environment variables, then the behavior is dependent on your shell. On the other hand, if you set them through arguments, the behavior is dependent on the `configure` script, which is more portable.

JaeseungYeom: Here you are hard-coding the compiler. to gcc/g++ in the path. Some gcc in some linux distribution comes in a different name like gcc10-gcc/gcc10-g++. Eventually, we want to support clang as well.

ilumsden: I just pushed a commit that removes the explicit compiler setting. Since the flux-core Docker images (which we're using in this Action) just use the default compilers for building, it makes sense to do the same here for the time being. 

We can revisit this later once we are looking to extend with other compilers. Since that's going to take some overhauling, I think it makes sense to do this in a separate PR.

PR #39, opened 2023-07-28
Caliper Integration

ilumsden:
This PR adds annotations and support for [Caliper](https://software.llnl.gov/Caliper/index.html) to DYAD.

This is still WIP because I need to add code to `configure.ac` to locate Caliper.

Once that code is added, users can enable Caliper annotations at build time by adding `--enable-caliper` to the invocation of the `configure` script.

At runtime, users can enable Caliper using the `CALI_CONFIG` environment variable (as is standard for Caliper).


ilumsden:
This PR is based on #24.

ilumsden:
Closing this PR in favor of a newer and better implementation

PR #41, opened 2023-08-11
fix CI

hariharan-devarajan:


hariharan-devarajan:
@JaeseungYeom I was trying to fix the CI so that we can make sure we have compilation test as well as some simple unit tests. 

Looks like only two versions of flux-compile work. Also can you point me to build a test case for testing DYAD functionality?

PR #49, opened 2023-10-05
Makes UCX an optional dependency

ilumsden:
This PR reworks `configure.ac` and all the relevant `Makefile.am`s to make UCX an optional dependency for DYAD.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -66,14 +66,21 @@ AC_ARG_ENABLE([urpc],
     [enable_urpc=$enableval],
     [enable_urpc=no]
 )
-AC_ARG_ENABLE([perfflow],
-    [AS_HELP_STRING([--enable-perfflow],
+AM_CONDITIONAL([URPC], [test "x$enable_urpc" = "xyes"])
+AC_ARG_WITH([perfflow],
+    [AS_HELP_STRING([--with-perfflow],
                     [enable performance measurement with PerfFlow Aspect])],
-    [enable_perfflow=$enableval],
-    [enable_perfflow=no]
+    [with_perfflow=$withval],
+    [with_perfflow=no]
 )
-# TODO Add support for libb64 back once base64 encoding/decoding is fully complete
-# AC_ARG_VAR([LIBB64_DIR], [root directory for libb64])
+AM_CONDITIONAL([PERFFLOW], [test "x$with_perfflow" = "xyes"])
+AC_ARG_WITH([ucx],
+    [AS_HELP_STRING([--with-ucx],
+                    [build the UCX-based data transport layer])],
+    [with_ucx=$withval],
+    [with_ucx=no]
+)
+AM_CONDITIONAL([UCX], [test "x$with_ucx" = "xyes"])
 
 #############################################
 # Define PKG_CHECK_VAR if it does not exist #
@@ -102,11 +109,29 @@ AS_VAR_IF([$1], [""], [$5], [$4])dnl
 ########################
 # Check for the dl library (specifically, the "dlsym" function)
 AC_CHECK_LIB([dl], [dlsym])
-# Check for UCX v1.6.0 or higher (Required)
-# TODO: make UCX optional based on a new AC_ARG_ENABLE flag
-PKG_CHECK_MODULES([UCX],
-    [ucx >= 1.6.0]
+# Find and get info for Flux-Core using pkg-config
+AX_FLUX_CORE
+# Check for Jansson 2.10 or higher (Required)
+PKG_CHECK_MODULES([JANSSON],
+    [jansson >= 2.10]
 )
+# Check for PerfFlow Aspect
+AX_PERFFLOW_ASPECT([PERFFLOW],
+    [pkg_check_perfflow_found=yes],
+    [pkg_check_perfflow_found=no]
+)
+if test "x$with_perfflow" = "xyes" && test "x$pkg_check_perfflow_found" = "xno"; then
+    AC_MSG_ERROR([requested PerfFlow Aspect support, but cannot find PerfFlow Aspect with pkg-config])
+fi
+# Check for UCX v1.6.0 or higher
+PKG_CHECK_MODULES([UCX],
+    [ucx >= 1.6.0],
+    [pkg_check_ucx_found=yes],
+    [pkg_check_ucx_found=no]
+)
+if test "x$with_ucx" = "xyes" && test "x$pkg_check_ucx_found" = "xno"; then
+    AC_MSG_ERROR([requested UCX support, but cannot find UCX with pkg-config])
+fi
 PKG_CHECK_VAR([UCX_LIBDIR],
     [ucx >= 1.6.0],
     [libdir],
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -66,14 +66,21 @@ AC_ARG_ENABLE([urpc],
     [enable_urpc=$enableval],
     [enable_urpc=no]
 )
-AC_ARG_ENABLE([perfflow],
-    [AS_HELP_STRING([--enable-perfflow],
+AM_CONDITIONAL([URPC], [test "x$enable_urpc" = "xyes"])
```
hariharan-devarajan, line 69: The AM_CONDITIONAL is not matching with AC_ARG_WITH. Fix the order to keep them together.

configure.ac
```diff
@@ -66,14 +66,21 @@ AC_ARG_ENABLE([urpc],
...
+AM_CONDITIONAL([PERFFLOW], [test "x$with_perfflow" = "xyes"])
+AC_ARG_WITH([ucx],
+    [AS_HELP_STRING([--with-ucx],
+                    [build the UCX-based data transport layer])],
+    [with_ucx=$withval],
+    [with_ucx=no]
 )
-# TODO Add support for libb64 back once base64 encoding/decoding is fully complete
-# AC_ARG_VAR([LIBB64_DIR], [root directory for libb64])
+AM_CONDITIONAL([UCX], [test "x$with_ucx" = "xyes"])
```
hariharan-devarajan, line 83: What is the purpose of having both with_ucx and enable ucx both having yes/no values. 

Generally --with-ucx should take the path right? and --enable-ucx would be just enabling it.

configure.ac
```diff
@@ -102,11 +109,29 @@ AS_VAR_IF([$1], [""], [$5], [$4])dnl
...
+    AC_MSG_ERROR([requested PerfFlow Aspect support, but cannot find PerfFlow Aspect with pkg-config])
+fi
+# Check for UCX v1.6.0 or higher
 PKG_CHECK_MODULES([UCX],
-    [ucx >= 1.6.0]
+    [ucx >= 1.6.0],
+    [pkg_check_ucx_found=yes],
+    [pkg_check_ucx_found=no]
 )
+if test "x$with_ucx" = "xyes" && test "x$pkg_check_ucx_found" = "xno"; then
```
hariharan-devarajan, line 132: should this be with_ucx or with enable-ucx? with-ucx is generally optional only to add when we don't know the path to its pkgconfig. I think enable-ucx should be the main check.

configure.ac
```diff
@@ -116,25 +141,11 @@ PKG_CHECK_VAR([UCX_LIBDIR],
...
-fi
-AM_CONDITIONAL([URPC], [test "x$enable_urpc" = "xyes"])
-AX_PERFFLOW_ASPECT([PERFFLOW],
-    [pkg_check_perfflow_found=yes],
-    [pkg_check_perfflow_found=no]
+AS_IF([test "x$with_ucx" = "xyes"],
+    [DYAD_MOD_RPATH="-Wl,-rpath,$UCX_LIBDIR"],
+    [DYAD_MOD_RPATH=""]
 )
-if test "x$enable_perfflow" = "xyes" && test "x$pkg_check_perfflow_found" = "xno"; then
```
hariharan-devarajan, line 134: This got changed to with_perfflow. I would recommend only adding or removing things for UCX in this PR.

src/core/Makefile.am
```diff
@@ -3,23 +3,22 @@ libdyad_core_la_SOURCES = \
...
 	-I$(top_srcdir)/src/utils/base64 \
 	-I$(top_srcdir)/src/dtl \
-    $(UCX_CFLAGS) \
     $(JANSSON_CFLAGS) \
 	$(FLUX_CORE_CFLAGS) \
 	-DBUILDING_DYAD=1 \
 	-fvisibility=hidden
 libdyad_core_la_CPPFLAGS = 
-libdyad_core_la_LDFLAGS = \
-	-Wl,-rpath,'$(UCX_LIBDIR)' \
```
hariharan-devarajan, line 21: do u wanna add UCX_LIBDIR? we have to make sure its pkgconfig correctly sets these.

What about rpath? skipping this will make libs dynamically loaded based on LD_LIBRARY_PATH.

src/core/Makefile.am
```diff
@@ -3,23 +3,22 @@ libdyad_core_la_SOURCES = \
...
-	$(AM_LDFLAGS)
+if UCX
+libdyad_core_la_LIBADD += $(UCX_LIBS)
+libdyad_core_la_CFLAGS += $(UCX_CFLAGS)
```
hariharan-devarajan, line 20: What about CPPFLAGS?

src/modules/Makefile.am
```diff
@@ -1,31 +1,33 @@
...
+# Don't put a line break before DYAD_MOD_RPATH in case it evaluates to an empty string
 dyad_la_LDFLAGS = \
 	$(AM_LDFLAGS) \
 	-module \
 	-avoid-version \
 	-no-undefined \
 	--disable-static \
 	-shared \
-	-export-dynamic \
-    -Wl,-rpath,'$(UCX_LIBDIR)'
```
hariharan-devarajan, line 12: Again same comment for -rpath and UCX_LIBDIR

hariharan-devarajan (changes requested):
I have a few comments that should be addressed.

src/core/Makefile.am
ilumsden: I've already checked that UCX's pkgconfig sets `libdir`. It works as expected.

As for your question about `rpath`, I have removed that flag from `libdyad_core.so`. 

The `rpath` flag is now only set for the module (i.e., `dyad.so`). This seemingly has to be done for the module due to some oddities when launching it with `flux module load`. For some reason, I've found that no environment variables from the user's environment are carried over into the launch of the module. So, `LD_LIBRARY_PATH` does nothing to the module. This is a problem for us because UCX has to be loaded for the module to work. By setting `rpath` for the module, I can get around this oddity.

src/core/Makefile.am
ilumsden: We only get `CFLAGS` from pkgconfig.

configure.ac
ilumsden: I'm not sure what you mean here. Unless GitHub did something dumb when I pushed, the `AM_CONDITIONAL` immediately follows the relevant `AC_ARG_<ENABLE|WITH>`

configure.ac
ilumsden: I'm not sure what you mean. There is no `--enable-ucx` in `configure.ac`

configure.ac
ilumsden: I chose with-ucx because, based on the GNU docs, with-<pkg> is used when dealing with optional dependencies. enable-<feature> is used to enable features.

configure.ac
ilumsden: For the time being, I just changed this to `with_perfflow`. If we want to revert that, it's a simple fix

configure.ac
hariharan-devarajan: I know generally u wanna say u have an option called --enable-ucx 
and optionally u can set path to ucx with an option called --with-ucx

ilumsden: I see. That makes more sense. I missed that characteristic of `with-package`. I'll switch to `AC_ARG_ENABLE` for these. I can add `AC_ARG_WITH` before taking the PR out of draft.

configure.ac
hariharan-devarajan: Oh i get it. Its all one after the other. I think for better code readability add a new line with a comment saying adding UCX/URPC/PERFLOW OPTIONS.

configure.ac
hariharan-devarajan: To me ucx is a feature we added right?

configure.ac
hariharan-devarajan: I am ok with it. but enable and with are different things.

src/core/Makefile.am
hariharan-devarajan: Thats fine.

hariharan-devarajan (approved):
This looks ok.

JaeseungYeom (approved):
LGTM

PR #50, opened 2023-10-05
WIP: Changes the tag generation for the UCX DTL

### LINKED ISSUE #34: Change UCX tag generation to be more unique per communication
Yet another suggestion from @hariharan-devarajan

Currently, the tags for UCX's tag-matching communication are generated with:

```
tag = (producer_rank << 32) | consumer_rank
```

The benefit of this approach is it makes use of information from Flux to create a tag unique to the producer and consumer applications. However, the downside is that multiple communications between the same producer/consumer pair will not use a unique tag, which introduces potential issues with UCX tag matching.

Instead of the current approach, we could maintain a 32-bit counter (or possibly larger) in the module. Using this counter, the module could generate tags using:

```
tag = (producer_rank << 32) | counter
```

This tag will be unique for a particular transfer that is outgoing from a particular producer module. As a result, all transfers will have a unique tag until the counter overflows.

ilumsden:
Closes #34 

This PR changes the tag generation for UCX tag matching. The new tags will be unique for every communication.

This is a work-in-progress PR. This feature is fully implemented, but it will need to be tested before merge.

### DIFF
--- a/src/dtl/ucx_dtl.c
+++ b/src/dtl/ucx_dtl.c
@@ -102,10 +102,6 @@ static ucs_status_t dyad_ucx_request_wait (dyad_dtl_ucx_t* dtl_handle,
     return UCS_OK;
 }
 
-static inline dyad_rc_t dyad_dtl_ucx_finalize_impl (dyad_dtl_ucx_t** dtl_handle)
-{
-}
-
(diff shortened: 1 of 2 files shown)


ilumsden:
Check if this is still applicable after other PRs are merged

PR #55, opened 2023-11-08
Integrate DLIO Benchmark with dyad's python APIs.

hariharan-devarajan:


tests/integration/dlio_benchmark/dyad_torch_data_loader.py
```diff
@@ -0,0 +1,126 @@
...
+    def __len__(self):
+        return self.num_samples
+
+    @dlp.log
+    def __getitem__(self, image_idx):
+        self.num_images_read += 1
+        step = int(math.ceil(self.num_images_read / self.batch_size))
+        logging.info(f"{utcnow()} Rank {get_rank()} reading {image_idx} sample")
+        filename, sample_index = self.conf.global_index_map[image_idx]
+        base_fname = os.path.basename(filename)
```
hariharan-devarajan, line 56: prepend with MANAGED DIRECTORY

hariharan-devarajan:
@JaeseungYeom I can confirm that the changes work locally. On the CI the call to dyad_open fails and then it gets stuck. @ilumsden Can u please look at the ci and suggest what can be done?

hariharan-devarajan:
@JaeseungYeom The current state is that i think the integration is fine, but the ci keeps hanging. We need more investigation + logging + timeouts to fix these kinds of issues.

JaeseungYeom (approved):
LGTM

PR #57, opened 2023-12-03
Changes to support mmap reads on Broker

hariharan-devarajan:
Changes to support mmaped reads on broker.

### DIFF
--- a/src/utils/read_all.c
+++ b/src/utils/read_all.c
@@ -21,6 +21,12 @@
 #include <sys/mman.h>
 #include "read_all.h"
 
+void * allocate_aligned(size_t size) {
+  size_t alignment = getpagesize();
+  int malloc_offset = (size + alignment) % alignment; 
+  void* mem = malloc(size + alignment);
+  return (char*) mem + malloc_offset;
+}
 ssize_t write_all (int fd, const void *buf, size_t len)
 {
     ssize_t n = 0;
@@ -57,7 +63,7 @@ ssize_t read_all (const char* filename, void **bufp)
         errno = EINVAL;
         return -1;
     }
-    *bufp = malloc (file_size);
+    *bufp = allocate_aligned (file_size);
     if (*bufp == NULL) {
         errno = EINVAL;
         return -1;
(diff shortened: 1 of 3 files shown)


src/utils/read_all.c
```diff
@@ -11,13 +11,22 @@
...
+#include <fcntl.h>
+#include <sys/mman.h>
 #include "read_all.h"
 
+void * allocate_aligned(size_t size) {
+  size_t alignment = getpagesize();
+  int malloc_offset = (size + alignment) % alignment; 
+  void* mem = malloc(size + alignment);
+  return (char*) mem + malloc_offset;
+}
```
JaeseungYeom, line 29: I do not think that this will guarantee that the address is a multiple of alignment, while the size of the usable block is. In addition, i am wondering if it would be safe to use such an address with munmap().

src/utils/read_all.c
```diff
@@ -37,8 +46,13 @@ ssize_t write_all (int fd, const void *buf, size_t len)
     return count;
 }
 
-ssize_t read_all (int fd, void **bufp)
+ssize_t read_all (const char* filename, void **bufp)
```
JaeseungYeom, line 49: This interface probably should take a page aligned memory address as an input for the UCX managed memory. So, the allocation would happen outside of this function.

JaeseungYeom (commented):
I left some comments but I know you will be on vacation. So, I can make changes if you agree with my comments.

src/utils/read_all.c
hariharan-devarajan: That is the buffer bufp. Once we do the UCX, we can move allocation there.

hariharan-devarajan:
@JaeseungYeom do u want me to work on this?

ilumsden:
Need to check the following before continuing:
- Is UCX-allocated memory page aligned?
- Is UCX-allocated memory with POSIX `mmap`?
- For client (i.e., producer/consumer), is UCX compatible with GPU memory allocators (e.g., `hipMalloc` on El Cap)?

PR #65, opened 2024-01-18
Build: cmake enhancement

JaeseungYeom:
1. Add compiler specific cmake controls
2. Add config header generation to avoid passing compiler flags via -D
3. GNUInstallDir based path
4. Option to promote compilation warnings to errors

For some reason, I see some files missing under src/dyad/stream.=> I see those under include/dyad


JaeseungYeom:
Checks are failing due to the compile-time generated header (dyad/dyad_config.h) is not found for some reason.

hariharan-devarajan (approved):

PR #71, opened 2024-01-20
Test: fix some issues with python API test

JaeseungYeom:
tests/pydyad_spsc seems to have multiple issues.

- `KVS_NAMESPACE` variable is inconsistently used with `DYAD_KVS_NAMESPACE`
- DYAD managed directory is not created.
- In case that such a directory already exists, and the directory contains the files from the previous runs, the test is invalid
- module load uses producer's managed path which is misleading due to the variable name
- Depending on the DTL mode, it shows different errors

### DIFF
--- a/.github/workflows/compile_test.yaml
+++ b/.github/workflows/compile_test.yaml
@@ -292,9 +292,9 @@ jobs:
       - name: Test DYAD with separate FS python
         run: |
           mkdir -p $DYAD_PATH
-          #. ${SPACK_DIR}/share/spack/setup-env.sh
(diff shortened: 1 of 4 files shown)


JaeseungYeom:
I test it manually using 2 compute nodes. It was working for both UCX and FLUX_RPC.

hariharan-devarajan:
@JaeseungYeom the ci seem to have no such file or directory. Do we need to create any directories on ci for that

hariharan-devarajan (approved):
The ci is still not working otherwise the changes look good to me.

JaeseungYeom:
I think the CI failure is a separate issue. We should look at the memory usage. I will add something and see how much memory it uses.

hariharan-devarajan:
Ok if that's the case I can merge it.

PR #58, opened 2023-12-13
Adds optimizations to the UCX DTL to improve performance

ilumsden:
This PR adds optimizations to the UCX DTL to improve DYAD's performance. Currently, the optimizations that will be added are:
1. Allocating memory buffers using UCX
2. Running a warmup communication with self to help initialize UCX components
3. Cache UCX endpoints in server to minimize the number of connection establishments required

PR #73, opened 2024-01-21
DTL merge conflct resolution

JaeseungYeom:
PR #58 has been diverged from the main branch due to recent CMake-based refactoring as well as file locking and reinitialization features.

This PR is to resolve the merge conflicts.


src/dyad/common/dyad_flux_log.h
```diff
@@ -7,15 +7,57 @@
...
+     fflush (stdout))
+#define DYAD_LOG_ERR(dyad_ctx, ...)  \
+    (fprintf (stderr, __VA_ARGS__),  \
+     fflush (stderr))
+#define FLUX_LOG_INFO(flux_ctx, ...) \
+    (fprintf (stdout, __VA_ARGS__),  \
+     fflush (stdout))
+#define FLUX_LOG_ERR(flux_ctx, ...)  \
+    (fprintf (stderr, __VA_ARGS__),  \
+     fflush (stderr))
```
JaeseungYeom, line 53: Multiple statements substituted by a macro should be grouped as one. Otherwise, it can be misused.
e.g.,
```
#MACRO_FUNC() statement1; statement2

if (CHK)
   MACRO_FUNC();
```
There are multiple options. I tried "(st1, st2)" option here but if this does not work try "do {st1; st2} while(0)"

src/dyad/core/dyad_core.c
```diff
@@ -4,17 +4,17 @@
...
-
 #include <dyad/core/dyad_envs.h>
 #include <dyad/dtl/dyad_dtl_impl.h>
-#include <dyad/common/dyad_flux_log.h>
+#include <dyad/perf/dyad_perf.h>
 #include <dyad/utils/murmur3.h>
 #include <dyad/utils/utils.h>
+#include <libgen.h>
+#include <unistd.h>
+#include <fcntl.h>
```
JaeseungYeom, line 16: I like the ordering of headers you are exercising here. including from specific to generic c headers.

src/dyad/core/dyad_core.c
```diff
@@ -57,16 +58,8 @@ static int gen_path_key (const char* str,
...
-                                       104683u,
-                                       104693u,
-                                       104701u,
-                                       104707u,
-                                       104711u,
-                                       104717u,
-                                       104723u,
-                                       104729u};
+    static const uint32_t seeds[10] =
+        {104677u, 104681u, 104683u, 104693u, 104701u, 104707u, 104711u, 104717u, 104723u, 104729u};
```
JaeseungYeom, line 62: I always have been wondering why you made this a multi-line statement or more than 2 lines. Now, it is better.

src/dyad/core/dyad_core.c
```diff
@@ -35,6 +35,7 @@
 const struct dyad_ctx dyad_ctx_default = {
     NULL,   // h
     NULL,   // dtl_handle
+    NULL,   // perf_handle
```
JaeseungYeom, line 38: will have to check if python API is up-to-date regarding this.

src/dyad/core/dyad_core.c
```diff
@@ -152,15 +153,16 @@ DYAD_CORE_FUNC_MODS dyad_rc_t publish_via_flux (const dyad_ctx_t* restrict ctx,
         goto publish_done;
     }
     rc = DYAD_RC_OK;
+kvs_transaction_create_and_pack_region_finish:;
```
JaeseungYeom, line 156: goto target lables should ends with :; instead of : to avoid compiler warning with certain compilers. I added ';'. through out the PR.

src/dyad/core/dyad_core.c
```diff
@@ -470,15 +474,17 @@ get_done:;
...
     DYAD_LOG_INFO (ctx, "Destroy the Flux future for the RPC\n");
     flux_future_destroy (f);
+    DYAD_PERF_REGION_END (ctx->perf_handle, "dyad_get_data");
     return rc;
 }
 
 DYAD_CORE_FUNC_MODS dyad_rc_t dyad_cons_store (const dyad_ctx_t* restrict ctx,
                                                const dyad_metadata_t* restrict mdata,
                                                int fd, const size_t data_len,
-                                               const char* restrict file_data)
```
JaeseungYeom, line 481: This should not be const here. I corrected.

src/dyad/core/dyad_core.c
```diff
@@ -955,18 +984,22 @@ dyad_rc_t dyad_consume (dyad_ctx_t* ctx, const char* fname)
...
-        free ((void*) file_data);
+        ctx->dtl_handle->return_buffer (ctx->dtl_handle, (void**)&file_data);
     }
     // Set reenter to true to allow additional intercepting
     ctx->reenter = true;
     return rc;
 }
 
-int dyad_finalize (dyad_ctx_t** ctx)
+dyad_rc_t dyad_finalize (dyad_ctx_t** ctx)
```
JaeseungYeom, line 994: good catch

src/dyad/core/dyad_core.c
```diff
@@ -516,7 +523,8 @@ DYAD_CORE_FUNC_MODS dyad_rc_t dyad_cons_store (const dyad_ctx_t* restrict ctx,
     }
     rc = DYAD_RC_OK;
 
-pull_done:
+pull_done:;
```
JaeseungYeom, line 526: We restructured consumer core code to use file locking. file_data is deallocated at consume_done rather than pull_done

src/dyad/core/dyad_core.h
```diff
@@ -35,26 +35,29 @@
...
-    uint32_t rank;                // Flux rank for DYAD
-    uint32_t service_mux;         // Number of Flux brokers sharing node-local storage
-    uint32_t node_idx;            // Index of the node hosting broker(s)
-    int pid;                      // unix process id, obtained by getpid()
-    char* kvs_namespace;          // Flux KVS namespace for DYAD
-    char* prod_managed_path;      // producer path managed by DYAD
-    char* cons_managed_path;      // consumer path managed by DYAD
+    flux_t* h;                      // the Flux handle for DYAD
+    struct dyad_dtl* dtl_handle;    // Opaque handle to DTL info
+    struct dyad_perf* perf_handle;  // Opaque handle to performance measurement tools
```
JaeseungYeom, line 46: Added by Ian

src/dyad/core/dyad_core.h
```diff
@@ -35,26 +35,29 @@
...
+    bool check;                     // if true, perform some check logging
+    bool reenter;                   // if false, do not recursively enter DYAD
+    bool initialized;               // if true, DYAD is initialized
+    bool shared_storage;            // if true, the managed path is shared
+    unsigned int key_depth;         // Depth of bins for the Flux KVS
+    unsigned int key_bins;          // Number of bins for the Flux KVS
+    uint32_t rank;                  // Flux rank for DYAD
+    uint32_t service_mux;           // Number of Flux brokers sharing node-local storage
+    uint32_t node_idx;              // Index of the node hosting broker(s)
+    int pid;                        // unix process id, obtained by getpid()
```
JaeseungYeom, line 57: Added by me just FYI

src/dyad/core/dyad_core.c
```diff
@@ -427,7 +432,7 @@ DYAD_CORE_FUNC_MODS dyad_rc_t dyad_get_data (const dyad_ctx_t* ctx,
         goto get_done;
     }
     DYAD_LOG_INFO (ctx, "Establish DTL connection with DYAD module");
-    rc = ctx->dtl_handle->establish_connection (ctx->dtl_handle, DYAD_COMM_RECV);
+    rc = ctx->dtl_handle->establish_connection (ctx->dtl_handle);
```
JaeseungYeom, line 435: This is the major change. With the new DTL interface, DYAD_COMM_RECV is set at initialization. So, it does not need to be passed here.

src/dyad/core/dyad_core.c
```diff
@@ -608,24 +629,26 @@ dyad_rc_t dyad_init (bool debug,
...
-        return DYAD_RC_NOCTX;
+        rc = DYAD_RC_NOCTX;
+        goto init_region_finish;
     }
     strncpy ((*ctx)->kvs_namespace, kvs_namespace, namespace_len + 1);
     // Initialize the DTL based on the value of dtl_mode
     // If an error occurs, log it and return an error
     FLUX_LOG_INFO ((*ctx)->h, "DYAD_CORE: inintializing DYAD DTL");
-    rc = dyad_dtl_init (&(*ctx)->dtl_handle, dtl_mode, (*ctx)->h, (*ctx)->debug);
+    rc = dyad_dtl_init (&(*ctx)->dtl_handle, dtl_mode, DYAD_COMM_RECV, (*ctx)->h, (*ctx)->debug, (*ctx)->perf_handle);
```
JaeseungYeom, line 648: Initialization with the extra argument, DYAD_COMM_RECV, of the new API.

src/dyad/modules/dyad.c
```diff
@@ -164,21 +185,32 @@ dyad_fetch_request_cb (flux_t *h, flux_msg_handler_t *w, const flux_msg_t *msg,
...
+        DYAD_PERF_REGION_END (ctx->perf_handle, "dyad_server_read_data");
         goto fetch_error;
     }
-    if ((inlen = read_all (fd, &inbuf)) < 0) {
+    FLUX_LOG_INFO (h, "Getting file size");
+    file_size = get_file_size (fd);
+    FLUX_LOG_INFO (h, "Getting buffer for reading the file");
+    rc = ctx->dtl_handle->get_buffer (ctx->dtl_handle, file_size, &inbuf);
+    FLUX_LOG_INFO (h, "Reading file data into buffer");
+    if ((inlen = read (fd, inbuf, file_size)) < 0) {
```
JaeseungYeom, line 201: I change the call to your version of `read_all()` to `read()`. So, there is no need to change `read_all.{h,c}`
In fact, the error semantic of the function was incorrect.

src/dyad/common/dyad_flux_log.h
```diff
@@ -7,15 +7,57 @@
...
 #endif
 
+// clang-format off
 #include <sys/types.h>
-#include <unistd.h>
+#ifdef __cplusplus
+#include <cstdio>
+#else
+#include <stdio.h>
+#endif
```
JaeseungYeom, line 16: replaced `#include <stdio.h>` with these lines

JaeseungYeom (commented):
Take a look.

JaeseungYeom:
Will need to find a way not to explicitly pass the perf handle.
Profiling should avoid adding extra overhead as much as possible however small that will be.

JaeseungYeom:
I had to change `perf_handle` to `self->perf_handle` in line 812 of `src/dyad/dtl/ucx_dtl.c` to build.

hariharan-devarajan:
@JaeseungYeom why can't we use DYAD ctx everywhere and have these kinds of optional args in there?

hariharan-devarajan:
The way we do signature for logging and profiling it is not very easy to extend. I feel we should have an internal ctx to functions and pass it to all functions. Including logging and profiling

JaeseungYeom:
> The way we do signature for logging and profiling it is not very easy to extend. I feel we should have an internal ctx to functions and pass it to all functions. Including logging and profiling

We can discuss this All those user facing options are directly used by the runtime. But there will be things that are used by the runtime while not being visible to users.

JaeseungYeom:
@hariharan-devarajan is working on fixing the bug in module's argparse.

- The header is indirectly included from perf. Perf should not be included when profiling is off.
- "flux/optparse.h" does not seem portable. I think we either embed the CRAN code as base64 or use a standard library. e.g., C getopt or boost program_options

PR #77, opened 2024-01-24
Consume after kvs pull

JaeseungYeom:
1. Fixes a bug in `dyad_consume()` that returns the buffer that has never been assigned.
2. Adds an variant interface of `dyad_consume()` that relies on the metadata that has already been pulled from KVS 
3. Still need to add a python interface: note that the new interface `dyad_consume_w_mdata(..., mdata)` deallocates mdata internally. We can change this behavior depending on how we want to write python interface.

### DIFF
--- a/src/dyad/core/dyad_core.c
+++ b/src/dyad/core/dyad_core.c
@@ -861,7 +861,7 @@ dyad_rc_t dyad_get_metadata (dyad_ctx_t* ctx,
     DYAD_C_FUNCTION_UPDATE_STR ("upath", upath);
     DYAD_LOG_INFO (ctx, "Generating KVS key: %s", topic);
     gen_path_key (upath, topic, topic_len, ctx->key_depth, ctx->key_bins);
-    rc = dyad_kvs_read (ctx, topic, fname, should_wait, mdata);
+    rc = dyad_kvs_read (ctx, topic, upath, should_wait, mdata);
     if (DYAD_IS_ERROR (rc)) {
         DYAD_LOG_ERROR (ctx, "Could not read data from the KVS");
         goto get_metadata_done;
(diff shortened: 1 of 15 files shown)


src/dyad/core/dyad_core.c
```diff
@@ -303,6 +303,10 @@ DYAD_CORE_FUNC_MODS dyad_rc_t dyad_fetch (const dyad_ctx_t* restrict ctx,
     char upath[PATH_MAX] = {'\0'};
     const size_t topic_len = PATH_MAX;
     char topic[PATH_MAX + 1] = {'\0'};
+    if (ctx->shared_storage) {
+        rc = DYAD_RC_OK;
```
JaeseungYeom, line 307: I did not add `DYAD_C_FUNCTION_UPDATE_INT ("is_local", 1);` here to avoid adding obvious information while increasing the log size.

hariharan-devarajan: I think it useful to have to easily filter those out. :)

hariharan-devarajan (changes requested):
We should not duplicate the reading part. We should just split the functionality and call the same function. 

Have consume_data do the I/O
consume can do fetch of metadata and call consume_data

src/dyad/core/dyad_core.c
JaeseungYeom: When shared_storage is on all accesses are going to be treated as local. Would it be useful to have a single message at the beginning, which we may already have, about shared storage?

JaeseungYeom:
> We should not duplicate the reading part. We should just split the functionality and call the same function.
> 
> Have consume_data do the I/O consume can do fetch of metadata and call consume_data

I understand it will make it more consistent. However, it will also needlessly allow redundant KVS traffic even with C/C++ when python is not in use. The original form is the best, and addition of the two calls `dyad_get_metadata()` and `dyad_consume_w_mdata()` is a compromise for python. Do we want to apply that to C and C++?

JaeseungYeom:
What is your opinion on the meta_data deallocation?

JaeseungYeom:
- Removed the deallocation of mdata inside of the `consume_w_metadata()`.
- Made the mdata pointer argument const 
- Added python binding as we discussed.

hariharan-devarajan (approved):

hariharan-devarajan (approved):
Changes look good.

PR #78, opened 2024-01-26
Check local cache before doing kvs lookup.

hariharan-devarajan:


JaeseungYeom (approved):
LGTM.
However, we will have to get back how to better handle or rename owner_rank.
Here, it is not the owner of the original file but the copy in cache.

PR #81, opened 2024-01-29
Bug Fix for shared storage 

hariharan-devarajan:
In case of shared storage, we should not use KVS to synchronize Producer and Consumer. Instead, we first use FS locks to do the sync, if that fails (say consumer got the lock first), we fall back to the case of KVS for sync. 

This allows us to optimize our KVS calls by calling it only in case, locks looks the order battle ;).

Note, if use streams, there is no portable way to do locks. Therefore, this PR will use KVS only for that case.

- [x] Test it with collocation test from Ian.

### DIFF
--- a/src/dyad/common/dyad_structures.h
+++ b/src/dyad/common/dyad_structures.h
@@ -20,6 +20,7 @@ struct dyad_ctx {
     flux_t* h;                      // the Flux handle for DYAD
     struct dyad_dtl* dtl_handle;    // Opaque handle to DTL info
     const char* fname;              // Used to track which file is getting processed.
+    bool use_fs_locks;              // Used to track if fs locks should be used.
     // User Facing
     bool debug;                     // if true, perform debug logging
     bool check;                     // if true, perform some check logging
--- a/src/dyad/core/dyad_core.c
+++ b/src/dyad/core/dyad_core.c
@@ -35,6 +35,7 @@ const struct dyad_ctx dyad_ctx_default = {
     NULL,   // h
     NULL,   // dtl_handle
     NULL,   // fname
+    false,  // use_fs_locks
     false,  // debug
     false,  // check
     false,  // reenter
@@ -303,10 +304,6 @@ DYAD_CORE_FUNC_MODS dyad_rc_t dyad_fetch (const dyad_ctx_t* restrict ctx,
     char upath[PATH_MAX] = {'\0'};
     const size_t topic_len = PATH_MAX;
     char topic[PATH_MAX + 1] = {'\0'};
-    if (ctx->shared_storage) {
-        rc = DYAD_RC_OK;
-        goto fetch_done;
-    }
     memset (upath, 0, PATH_MAX);
     memset (topic, 0, topic_len + 1);
     // Extract the path to the file specified by fname relative to the
@@ -674,6 +671,7 @@ dyad_rc_t dyad_init (bool debug,
     // Set reenter and initialized to indicate this.
     (*ctx)->reenter = true;
     (*ctx)->initialized = true;
+    (*ctx)->use_fs_locks = true; // This is default value except for streams which dont have a fp.
     // TODO Print logging info
     rc = DYAD_RC_OK;
     // TODO: Add folder option here.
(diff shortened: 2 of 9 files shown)


JaeseungYeom:
I see the performance benefit for using file locking to synchronize between producer and consumer when they have access to the same storage. That is not a typical scenario intended for DYAD unless to deal with the hierarchical storage. We will have to bring back the hierarchical storage management related development once we are done with current publication efforts. However, the description of the bug is lacking here. If I understand the situation, this bug is introduced by the file locking optimization on client side. We no longer check KVS as the first method to determine the local availability of a file. Instead, we check if a file is locally available by opening it and putting a lock on it before checking KVS. If the producer has already opened a file but has not finished, the consumer will see a file and continue reading an incomplete file.

JaeseungYeom:
One solution could be to create a file (e.g., node_id_filename.lock) as a lock to indicate that writing is not complete before opening a file to write. Upon closing of the file, the producer removes the lock file. Consumer checks the existence of the local file instead of relying on the file size.

hariharan-devarajan:
> One solution could be to create a file (e.g., node_id_filename.lock) as a lock to indicate that writing is not complete before opening a file to write. Upon closing of the file, the producer removes the lock file. Consumer checks the existence of the local file instead of relying on the file size.

Another option is to do stream I/O with fd. This would open and write data into fd while the interface uses streams.

hariharan-devarajan:
> I see the performance benefit for using file locking to synchronize between producer and consumer when they have access to the same storage. That is not a typical scenario intended for DYAD unless to deal with the hierarchical storage. We will have to bring back the hierarchical storage management related development once we are done with current publication efforts. However, the description of the bug is lacking here. If I understand the situation, this bug is introduced by the file locking optimization on client side. We no longer check KVS as the first method to determine the local availability of a file. Instead, we check if a file is locally available by opening it and putting a lock on it before checking KVS. If the producer has already opened a file but has not finished, the consumer will see a file and continue reading an incomplete file. Checking if a file size is greater than 0 would not be sufficient. So, I think for a shared storage, we probably should use KVS only even on consumer side.

No, I added locks on producer side as well. So everything would be synchronized. For shared storage case with streams, as streams don't have locks we don't use locks to synchronize.

JaeseungYeom:
> > I see the performance benefit for using file locking to synchronize between producer and consumer when they have access to the same storage. That is not a typical scenario intended for DYAD unless to deal with the hierarchical storage. We will have to bring back the hierarchical storage management related development once we are done with current publication efforts. However, the description of the bug is lacking here. If I understand the situation, this bug is introduced by the file locking optimization on client side. We no longer check KVS as the first method to determine the local availability of a file. Instead, we check if a file is locally available by opening it and putting a lock on it before checking KVS. If the producer has already opened a file but has not finished, the consumer will see a file and continue reading an incomplete file. Checking if a file size is greater than 0 would not be sufficient. So, I think for a shared storage, we probably should use KVS only even on consumer side.
> 
> No, I added locks on producer side as well. So everything would be synchronized. For shared storage case with streams, as streams don't have locks we don't use locks to synchronize.

I see that now. Then I would like to go over again with the full context in mind. However, can you rebase?

hariharan-devarajan:
@JaeseungYeom I rebased it.

tests/integration/dlio_benchmark/setup-env.sh
```diff
@@ -32,7 +33,7 @@ export DLIO_PROFILER_LOG_LEVEL=ERROR
 export DLIO_PROFILER_BIND_SIGNALS=0
 export MV2_BCAST_HWLOC_TOPOLOGY=0
 
-
+mkdir -p ${DYAD_PATH}
```
JaeseungYeom, line 36: I think we should add -m 775 to be consistent with dyad core and module.

src/dyad/stream/dyad_stream_core.cpp
```diff
@@ -122,6 +122,7 @@ void dyad_stream_core::init (const dyad_params &p)
                               p.m_cons_managed_path.c_str (),
                               static_cast<dyad_dtl_mode_t> (p.m_dtl_mode),
                               &m_ctx);
+    m_ctx->use_fs_locks = false; // this is to disable checking for fs lock based logic.
```
JaeseungYeom, line 125: Add a TODO note saying that we will implement a lock file based synchronization scheme.

src/dyad/core/dyad_core.c
```diff
@@ -934,58 +931,71 @@ dyad_rc_t dyad_consume (dyad_ctx_t* ctx, const char* fname)
...
-            DYAD_LOG_INFO (ctx, "File '%s' is local!\n", fname);
-            rc = DYAD_RC_OK;
-            dyad_release_flock (ctx, fd, &exclusive_lock);
-            goto consume_close;
+    file_size = get_file_size (fd);
+    if (ctx->shared_storage) {
+        dyad_release_flock (ctx, fd, &exclusive_lock);
+        if (!ctx->use_fs_locks || file_size <= 0) {
+            // as file size was zero that means consumer won the lock first so has to wait for kvs.
+            // or we are using c++ streams.
```
JaeseungYeom, line 939: Clarify the comment by changing the following:
"or we are using c++ streams." => "or we cannot use file lock based synchronization as it does not work with the files managed by c++ fstream."

JaeseungYeom (approved):
If you make the changes I requested, it is ready for merging.

src/dyad/common/dyad_structures.h
```diff
@@ -20,6 +20,7 @@ struct dyad_ctx {
     flux_t* h;                      // the Flux handle for DYAD
     struct dyad_dtl* dtl_handle;    // Opaque handle to DTL info
     const char* fname;              // Used to track which file is getting processed.
+    bool use_fs_locks;              // Used to track if fs locks should be used.
```
JaeseungYeom, line 23: Make sure the compatibility with the python binding as well as the initialization of the struct in dyad_core.c

JaeseungYeom: 1. https://github.com/flux-framework/dyad/blob/6d3df1b5f680ca7295a1e46e7aa731a9982636e6/pydyad/pydyad/bindings.py#L24
2. https://github.com/flux-framework/dyad/blob/9a87cf0ac93b982fd0eac5857c0b175c9ce70f68/src/dyad/core/dyad_core.c#L38

JaeseungYeom:
I am currently doing something not so portable in another part of the code 
https://github.com/flux-framework/dyad/blob/6d3df1b5f680ca7295a1e46e7aa731a9982636e6/include/dyad/stream/dyad_stream_api.hpp#L65
based on:
https://stackoverflow.com/questions/676787/how-to-do-fsync-on-an-ofstream

If there is a way to make it portable, we can leverage that for locking.

hariharan-devarajan:
> I am currently doing something not so portable in another part of the code
> 
> https://github.com/flux-framework/dyad/blob/6d3df1b5f680ca7295a1e46e7aa731a9982636e6/include/dyad/stream/dyad_stream_api.hpp#L65
> 
> 
> based on:
> https://stackoverflow.com/questions/676787/how-to-do-fsync-on-an-ofstream
> This is a confession of my sin. If there is a way to make it portable, we can leverage that for locking.

I am not sure what you are doing here. But if u get a file descriptor, then u can add a similar logic to the open and close code I have for the producer case, and then it would work.

tests/integration/dlio_benchmark/setup-env.sh
hariharan-devarajan: FIXED

src/dyad/stream/dyad_stream_core.cpp
hariharan-devarajan: FIXED

src/dyad/core/dyad_core.c
hariharan-devarajan: FIXED

src/dyad/common/dyad_structures.h
hariharan-devarajan: Added

JaeseungYeom (approved):
LGTM

PR #82, opened 2024-01-31
Asynchronous KVS publish by producer enabled by DYAD_ASYNC_PUBLISH

JaeseungYeom:
WIP. We will test and see.
Set the environment variable DYAD_ASYNC_PUBLISH.


pydyad/pydyad/bindings.py
```diff
@@ -27,7 +27,7 @@ class DyadCtxWrapper(ctypes.Structure):
         ("reenter", ctypes.c_bool),
         ("initialized", ctypes.c_bool),
         ("shared_storage", ctypes.c_bool),
-        ("sync_started", ctypes.c_bool),
```
JaeseungYeom, line 30: I was not sure what this variable was for. So, I removed.

JaeseungYeom:
In the CI test, cpp_logger is not built with c11 standard and causes issue.

hariharan-devarajan (approved):
Lets wait for @ilumsden to test this, then we can merge.

ilumsden:
@JaeseungYeom caught a small bug in my testing. You correctly read `DYAD_ASYNC_PUBLISH` in `dyad_init_env`, but you never actually set the context's `async_publish` field in `dyad_init`. I just pushed a super tiny commit onto your branch to fix this.

ilumsden (approved):
Looks good to me. I also tested with the workflow benchmark, and this worked as expected.

PR #88, opened 2024-02-02
enable file locking for c++ fstream and ofstream. control fsync after…

JaeseungYeom:
1. enable file locking for c++ fstream and ofstream
3. control fsync after producer write via environment variable
4. change the order of lock release such that it comes just before closing a file locked.
5. change the function name dyad_fetch to dyad_fetch_metadata
6. added error check for dyad_fetch_metadata with shared_storage enabled
7. suppress warning about usleep


include/dyad/stream/dyad_stream_api.hpp
```diff
@@ -70,11 +71,17 @@ void fsync_ofstream (std::basic_ofstream<_CharT, _Traits>& os)
         os.flush ();
         fsync (static_cast<my_filebuf&> (*os.rdbuf ()).handle ());
     }
+#else
+    if (os.is_open ()) {
+        os.flush ();
```
hariharan-devarajan, line 76: Why do we need to do `os.flush()`?

include/dyad/stream/dyad_stream_api.hpp
```diff
@@ -88,8 +95,103 @@ void fsync_fstream (std::basic_fstream<_CharT, _Traits>& os)
...
+       public:
+        int handle ()
+        {
+            return this->_M_file.fd ();
+        }
+    };
+
+    if (os.is_open ()) {
+        int fd = static_cast<my_filebuf&> (*os.rdbuf ()).handle ();
+        core.file_lock (fd);
```
hariharan-devarajan, line 122: Is there a way to select which lock your applying? We should use shared locks for read and exclusive locks for writes,

hariharan-devarajan: ah i see u are only only applying for producer. I think, we should have different methods for shared or exclusive locks for clarity of API.

hariharan-devarajan (changes requested):

include/dyad/stream/dyad_stream_api.hpp
JaeseungYeom: To make sure that data is cleared out before starting the followup synchronization routine, which can happen before d'tor call.

include/dyad/stream/dyad_stream_api.hpp
hariharan-devarajan: I know maybe u wanna say exclusive lock

include/dyad/stream/dyad_stream_api.hpp
hariharan-devarajan: I would not do that always. As I said, we should avoid any kind of flush.

include/dyad/stream/dyad_stream_api.hpp
JaeseungYeom: We can add a shared lock to ifstream. Also to the fstream with `m_core.is_dyad_consumer()`

hariharan-devarajan: We don't need that as the consume code already does shared lock.

include/dyad/stream/dyad_stream_api.hpp
JaeseungYeom: For any fencing semantic, I think this is needed.

include/dyad/stream/dyad_stream_api.hpp
JaeseungYeom: I agree. That is why I  did not add it.

include/dyad/stream/dyad_stream_api.hpp
hariharan-devarajan: the session semantics is maintained by closing already, we don't need that for most file systems.

JaeseungYeom: This is decoupled from file close. This is a utility function like mem fence instruction. If you use it just before stream close, there will be nothing to flush at close.

hariharan-devarajan: So this wont be called automatically by close of stream? Users need to call it explicitly?

hariharan-devarajan: My understanding is that this would not be an expensive operation. Therefore its ok.

hariharan-devarajan (approved):
The changes look good now.

PR #91, opened 2024-02-03
Fix the broken logger, cleanup macros, fix warnings

JaeseungYeom:
- Fix the broken logger
  - Flux logger
  - Add directory hint for cpp-logger search in cmake
- cleanup macros
  - more structuring
  - noop handling
  - multiline expansion
- fix warnings
  - with clang, gcc, intel for c99, c11, c17, c++11, c++17


hariharan-devarajan (approved):
The changes look good to me.

PR #92, opened 2024-02-03
Optparse fix

JaeseungYeom:
This PR is based on PR #91 

- This is an attempt to fix the argument parsing in module. The modules argument parsing has been broken for a while. I tried various things to fix it but no success. So, I decided to do the complete overhaul by replacing it with `getopts_long`

- Added `-m 775` to mkdir in some scripts such that the permission is consistent with the expectation from dyad module and core.

This PR will still not pass the CI test. However, it is not because of the parsing issue but because of DTL initialization issue. I will fix it in a followup PR.


JaeseungYeom:
To actually see if DTL correctly initializes, need to run dtl_fix PR #93.
However, with this PR it is still possible to see the option argument parsing in the works. I added an offline test code. It is almost identical to the actual parsing in the module code. However, there is one key difference. In module, `argv[0]` is not the name of executable as the module is not an executable. So, arguments starts from `argv[0]` instead of `argv[1]`, which is different from what all the parsing library expect. So, I copy argv array into another array starting from the second element (`argv[1]`) to feed the parsing library.

hariharan-devarajan (approved):
Changes look good to me.

PR #93, opened 2024-02-03
Fix DTL initialization and build

JaeseungYeom:
When built for UCX_RMA, the initialization does not find UCX at all.
The CMake variable to enable UCX_RMA was not enabling any UCX dependency as well as UCX related initialization.
This RP is based on PR #92


hariharan-devarajan (approved):
The changes look good to me.

hariharan-devarajan (changes requested):
The CI is failing we should fix that but the changes look good.

PR #94, opened 2024-02-03
Move the DLIO test after Python API test

JaeseungYeom:
DLIO test requires Python API.
So, it makes sense to move it after Python API test.
We will also be able to see the test result with c core first.


JaeseungYeom:
This is a simple change in CI pipeline, which has no impact on the code base itself. To make it easy to diagnose the issue on other PRs, I am merging this.

PR #99, opened 2024-02-06
Improve consume performance by avoiding fsync.

### LINKED ISSUE #98: Fix DYAD consumer fsync
Currently as we have one fd we need an fsync before releasing a lock.

What we can do is have a dummy fd on which we take a lock. Then actual consumption (remote read + write) happens on a fd and we just close it. Then we release the lock on dummy fd and then close that fd as well. 

This would significantly improve I/O of consumer.

hariharan-devarajan:
We have two fd one for locking and one for io which we close so we dont have to do fsync.

Closes #98

### DIFF
--- a/src/dyad/core/dyad_core.c
+++ b/src/dyad/core/dyad_core.c
@@ -958,7 +958,7 @@ dyad_rc_t dyad_consume (dyad_ctx_t* ctx, const char* fname)
     DYAD_C_FUNCTION_START();
     DYAD_C_FUNCTION_UPDATE_STR ("fname", fname);
     dyad_rc_t rc = DYAD_RC_OK;
-    int fd = -1;
+    int lock_fd = -1, io_fd = -1;
     ssize_t file_size = -1;
     char* file_data = NULL;
     size_t data_len = 0ul;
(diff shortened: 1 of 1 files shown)


JaeseungYeom (approved):
I saw the `fsync ()` and remembered that we added it when we first worked on the file lock based synchronization. I was wondering whether we should have removed it in light of recent discussions.

src/dyad/core/dyad_core.c
```diff
@@ -979,21 +979,20 @@ dyad_rc_t dyad_consume (dyad_ctx_t* ctx, const char* fname)
     // Set reenter to false to avoid recursively performing
     // DYAD operations
     ctx->reenter = false;
-    fd = open (fname, O_RDWR | O_CREAT, 0666);
-    DYAD_C_FUNCTION_UPDATE_INT ("fd", fd);
-    if (fd == -1) {
+    lock_fd = open (fname, O_RDWR | O_CREAT, 0666);
```
JaeseungYeom, line 982: The last permission bit should be 0 or 4. So. 0660 or 0664

src/dyad/core/dyad_core.c
```diff
@@ -1030,41 +1029,45 @@ dyad_rc_t dyad_consume (dyad_ctx_t* ctx, const char* fname)
             rc = dyad_get_data (ctx, mdata, &file_data, &data_len);
             if (DYAD_IS_ERROR (rc)) {
                 DYAD_LOG_ERROR (ctx, "dyad_get_data failed!\n");
-                dyad_release_flock (ctx, fd, &exclusive_lock);
+                dyad_release_flock (ctx, lock_fd, &exclusive_lock);
                 goto consume_done;
             }
             DYAD_C_FUNCTION_UPDATE_INT ("data_len", data_len);
-
+            io_fd = open (fname, O_WRONLY);
```
JaeseungYeom, line 1036: I am not sure about this. The lock is released before the file is written. So, there is a chance that another process initiates the metadata full and data full before this file is written.

JaeseungYeom:
What is the reason to reopen the file in a write only mode?

PR #101, opened 2024-02-07
Residency

JaeseungYeom:
This PR depends on PR #91

This PR implements set-associative cache with LRU replacement scheme as well as basic priority-based replacement scheme.
This includes a test.


ilumsden:
Note: this PR adds a dependency on Boost

ilumsden:
Another option that we could use to support both `std::filesystem` and `std::experimental::filesystem`: https://github.com/vector-of-bool/CMakeCM/blob/master/modules/FindFilesystem.cmake

JaeseungYeom:
> Another option that we could use to support both `std::filesystem` and `std::experimental::filesystem`: https://github.com/vector-of-bool/CMakeCM/blob/master/modules/FindFilesystem.cmake
This PR does not use boost filesystem if that is what you are referring. The dependency is for boost::multi_index.
BTW, I think we had some cmake test for detecting which is available and then setting a macro. Based on that, you can choose which to use in your source code. In any case, std::filesystem is fully incorporated after c++17. std::experimental did not work that well for me. Either something is not working or something is missing.

wangvsa (approved):
Looks good to me. I will merge this PR first then integrate it with my side of code to implement the consumer-side caching. The integration may take some time and testing will be time-consuming as well.

PR #106, opened 2024-02-10
Singleton cpp ctx

JaeseungYeom:
Fixing  cpp api bug.
As @hariharan-devarajan pointed out, c'tor and d'tor of cpp dyad stream initializes and finalizes. As a result, ep-caching becomes stale.
So, a persistent singleton object is initialized only once and reused.

Actually, this PR is based on other PRs #100 , #102 and #104 , each of which fixes other bugs.


JaeseungYeom:
Both C and C++ are working on ruby but CI still fails.

PR #107, opened 2024-02-11
Check prefix first

JaeseungYeom:
This combines all the fixes plus the prefix check reorder.
This PR check the managed path prefix before locking any file. When a match is found, it sets the reenter flag to false to avoid dyad goes into recursive interception. This PR is base on PR #106


src/dyad/core/dyad_core.c
```diff
@@ -1116,17 +820,16 @@ dyad_rc_t dyad_consume_w_metadata (dyad_ctx_t* ctx, const char* fname,
     }
     // If the consumer-managed path is NULL or empty, then the context is not
     // valid for a consumer operation. So, return DYAD_BADMANAGEDPATH
-    if (ctx->cons_managed_path == NULL || strlen (ctx->cons_managed_path) == 0) {
+    if (ctx->cons_managed_path == NULL) {
```
JaeseungYeom, line 823: The comment above seems outdated?

JaeseungYeom, line 823: The comment at line 814-815 seem incorrect or outdated to me.

JaeseungYeom:
This now passes the checks.
However, I am not sure why we need the following lines
https://github.com/flux-framework/dyad/blob/a27f3f24aa10dfdc8368cf0ae5a9f1bd4053c648/src/dyad/core/dyad_core.c#L590-L614

Without that the check after the following comment might just work.
https://github.com/flux-framework/dyad/blob/a27f3f24aa10dfdc8368cf0ae5a9f1bd4053c648/src/dyad/core/dyad_core.c#L814

PR #109, opened 2024-02-12
improve hash distribution with short strings

JaeseungYeom:
Strings shorter than 128 bytes collide possible due to the implementation details of `MurmurHash3_x64_128()`. Especially, the hash value seems to depend on the length of such a string.
For such a short string, we concatenate it as many times as needed to make it longer than 128 bytes.
Added a test to show hash results.

For example, before this fix,  `test_murmur3 3 1024 data1.txt data2.txt data3.txt 123456.txt 1data5.txt data91.txt` showed the following results
```
data1.txt	199.ca.3e.data1.txt
data2.txt	199.ca.3e.data2.txt
data3.txt	199.ca.3e.data3.txt
123456.txt	37e.c8.89.123456.txt
1data5.txt	37e.c8.89.1data5.txt
data91.txt	37e.c8.89.data91.txt
```
After the fix, the below is the new results.
```
data1.txt	2cd.1cd.3cc.data1.txt
data2.txt	11f.2cb.5a.data2.txt
data3.txt	22b.28f.11d.data3.txt
123456.txt	db.272.14.123456.txt
1data5.txt	8a.1ce.f8.1data5.txt
data91.txt	219.32f.47.data91.txt
```

### DIFF
--- a/src/dyad/core/dyad_core.c
+++ b/src/dyad/core/dyad_core.c
@@ -49,6 +49,7 @@ static int gen_path_key (const char* restrict str,
     size_t cx = 0ul;
     int n = 0;
     size_t str_len = strlen (str);
+    const char* str_long = str;
 
     if (str == NULL || path_key == NULL || len == 0ul || str_len == 0ul) {
         DYAD_C_FUNCTION_END();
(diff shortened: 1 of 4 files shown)


hariharan-devarajan (approved):
The changes makes sense.

PR #113, opened 2024-02-28
change future destroy back to future reset in flux dtl recv

JaeseungYeom:
Fix the recv function of flux rpc based dtl.
This is to change the flux_future_destroy back to flux_future_reset for streaming RPC.
This is a bug I introduced some time ago.


ilumsden:
@hariharan-devarajan @ilumsden this is ready for review

PR #114, opened 2024-03-01
Changes to add unit test and Internal Testing

hariharan-devarajan:
1. Add unit test infrastructure with Catch
2. Fix a bug in ctx manager.
3. Added Test case for testing data transfer capability.
3. Added Test case for testing mdm capability.
4. Exported some functions for unit testing.


JaeseungYeom:
Can we distinguish tests that requires LC environment from general environment?

JaeseungYeom:
There are many scripts added. Please add comments at the top of each script describing what it is for, what arguments it takes, and what other scripts it depends on.

JaeseungYeom:
We will address the comments in the followup PRs.

PR #119, opened 2024-08-21
Switch release lock prints to Debug

hariharan-devarajan:
The current code keeps printing that the release lock happened in INFO which I think is higher level debugging. I have switched this to debug.


hariharan-devarajan:
@JaeseungYeom Lets finalize this. This is a small and simple PR :) .

PR #116, opened 2024-05-22
Completes the WIP CMake config for find_package

ilumsden:
The big thing this PR does is complete @JaeseungYeom's WIP `DYADConfig.cmake` and similar stuff for exporting DYAD and making it compatible with `find_package(DYAD CONFIG)`.

It also does the following more minor things:
* Makes the naming of install path variables (e.g., `DYAD_INSTALL_INCLUDE_DIR`) more consistent and tweaks the values to be more consistent with expected behavior for installation paths
* Replaces the glibc version check with a portable version
* Replaces the CMake targets that added `-Werror` with a macro that will directly add the flag to targets with `target_compile_options`

In case we want to revert any of these changes, this PR currently does not delete any of the old code. Instead, the old code is just all commented out.

These changes have all been tested with the unit testing from the in-progress refactor/enhancement of the A4MD benchmark. Thanks to that, I can confirm that these changes allow external projects to properly use DYAD as a dependency with `find_package`.

PR #120, opened 2024-09-03
Makes Core into Core and Client and Exposes Client API to Users

ilumsden:
This PR is based on feedback from #118. If we like it, it can supersede and replace #118.

This PR moves headers around and modifies the CMake to make DYAD's core and ctx libraries available to users. It also splits `dyad_core.h` and `dyad_structures.h` into multiple headers (an external header and an internal one) so that excess function definitions are not exposed to users. It also creates a new `dyad/core/dyad.h` header and `dyad_client` CMake target to refer to both libraries at once.


ilumsden:
I rebased this PR on top of #116 for testing purposes. Please merge #116 before merging this PR.

ilumsden:
I tested this using A4X, and it works as expected

wangvsa:
@ilumsden  is this ready to review/merge? This one and #124 are big PRs and I'd like to merge them first if possible.

ilumsden:
Yes. This PR is ready for review and merge.

PR #122, opened 2024-09-04
Removes all Autotools files and the redundant dyadrun.in file

### LINKED ISSUE #121: Remove old Autotools build system
Since moving from Autotools to CMake, we've made a large number of changes to the point that all the Autotools build files are invalid. At this point, we should remove the old Autotools stuff.

ilumsden:
Resolves #121.

This is a simple PR that just removes all the old, out-of-date Autotools stuff. It also removes the very redundant, out-of-date, and unnecessary `dyadrun.in` file.


hariharan-devarajan (approved):
It looks good to me.

PR #132, opened 2025-01-31
Allow lib64 as the dyad library installation path in test scripts

JaeseungYeom:
Currently, dyad library path is hard-coded as 'lib' in test scripts.
Allow 'lib64' as the dyad library installation path.


JaeseungYeom:
I am not sure what the error with readthedocs checkout failure. What is the revision shown in the error for "Failed to checkout revision: d5005197a07ca5ff346d7fbff056f0f0eb519d8c"? Why does it need to check it out that particular revision?

PR #134, opened 2025-03-23
build: allow look for pkgcfg of Flux core under lib64

JaeseungYeom:
cmake currently only look into lib to find flux_core package config. This PR allow it to find flux_core under lib64.


JaeseungYeom:
cmake failed to find flux_core library on corona because it is under /usr/lib64

PR #139, opened 2025-06-03
Bump OS and package versions

wangvsa:
The old Ubuntu is nog longer supported. Updated the system and many package versions.

### DIFF
--- a/pydyad/setup.cfg
+++ b/pydyad/setup.cfg
@@ -1,6 +1,8 @@
 [metadata]
 name = pydyad
 version = 0.1.0
+# Not supported. Comment it out
+# classifier = License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)
 
 [options]
 python_requires = >=3.7


.github/workflows/compile_test.yaml
```diff
@@ -275,29 +278,32 @@ jobs:
           cd ${GITHUB_WORKSPACE}/docs/demos/ecp_feb_2023
           make all
       - name: Test DYAD with separate FS c
-        timeout-minutes: 1
+        timeout-minutes: 2
```
JaeseungYeom, line 281: Was this necessary? If it needs two minutes to complete, something is going on.

JaeseungYeom: Can you run it with debug output enabled?

wangvsa: Fixed now. Have tested on my local repo.
https://github.com/wangvsa/dyad/actions/runs/15485596984/job/43599461514
We can merge this now. @JaeseungYeom

PR #142, opened 2025-06-18
Bug fix: pydyad has dftracer decorator on constructor which is before…

JaeseungYeom:
… dftracer initialization


JaeseungYeom:
1. We need a guard to avoid re-initializing DFTracer in `init` and `init_env` of pydyad binding.
2. Also, in dyad core, `dyad_init ()` is not supposed to be called directly. It is always supposed to be called from `dyad_init_env ()`. The only exception is in a c++ copy constructor. To avoid confusion, it would be better to change `init()` in pydyad to set env variables and call init_env rather than exposing `dyad_init ()` in the core.
3. I am still not clear how to handle the usage of dft for dyad_open. 
```@contextmanager
@dft_log.log
def dyad_open(*args, dyad_ctx=None, metadata_wrapper=None, register_dyad_ctx=False, **kwargs):
    global DYAD_IO
    local_dyad_io = dyad_ctx
    if dyad_ctx is None:
        if DYAD_IO is None:
            DYAD_IO = Dyad()
            DYAD_IO.init_env()
        local_dyad_io = DYAD_IO
    else:
        if register_dyad_ctx:
            DYAD_IO = dyad_ctx
     ....
```

PR #144, opened 2025-07-10
bug fix: fix out-of-scope access and pydyad test with PYTHONPATH

JaeseungYeom:
Two bugs fixed.
1) pydyad test fails because it cannot load pydyad module. Fixed it by adding pydyad path to PYTHONPATH
2) out-of-scope access. buf[] is defined within a scope and its pointer is assigned to a variable accessed out of the scope. Fixed it by moving the variable to the upper scope.


JaeseungYeom:
Can't find dftracer module.

PR #165, opened 2026-05-25
fetch_package

JaeseungYeom:
This PR is exclusively on cmake
- Replacing FindCppLogger and FindGotcha, which are cmake FetchContent wrappers with fetch_package utility function.
- Fixed outdated import config input files
- Changed the namespace from DYAD:: to dyad::

Currently waiting for the PRs to be merged at gotcha and cpp-logger repos.

- https://github.com/llnl/cpp-logger/pull/27#issue-4512592059
- https://github.com/llnl/GOTCHA/pull/172#issue-4512196902

Then, I will replace git url and tag used in fetch_package to the latest in each repo.

PR #166, opened 2026-05-26
Catch2 upgrade

JaeseungYeom:
This PR fixes outdated unit test relying on Catch2 v2. It depends on PR #165 

- Add FetchContent-based Catch2 dependency management.
- Fix outdated unit test relying on Catch2 v2 for new Catch2 v3.
- Add missing c++ guard in dyad_client_int.h to avoid linking error


JaeseungYeom:
Merging this to allow unit test to build

PR #170, opened 2026-06-01
Add comments

JaeseungYeom:
- Added a bunch of comments.
- Added Doxygen
  - Updated requirements.txt to include Breathe
  - Updated Makefile, config.py and index.rst
  - Created a folder docs/api and populated with *.rst files that integrate doxygen outcomes via breathe
  - Added a page explaining the use of file locking mechanism in DYAD
- Fixed a couple of bugs
   - dyad_stream_core::file_lock_shared was calling an exclusive lock. This interface has not been used. So, it should not have impacted any testing.
   - The fall back mechanism of c++ stream file locking (use_fs_lock==false) was not fully implemented. It is fixed by adding the condition check to dyad_consume() and initialization to dyad_wrapper_init(). We never had to deal with fallback situation. So, it should not have impacted any result.
   - There was a section of code in dyad_get_data() for UCX data transfer completion and was reached by FLUX RPC DTL case. Added a guard to prevent that.


src/dyad/client/dyad_client.c
```diff
@@ -456,17 +776,19 @@ get_done:;
...
-    ctx->dtl_handle->get_buffer (ctx, 0, (void **)file_data);
-    ssize_t read_len = 0l;
-    memcpy (&read_len, *file_data, sizeof (read_len));
-    if (read_len < 0l) {
-        *file_len = 0ul;
-        DYAD_LOG_DEBUG (ctx, "Not able to read from %s file", mdata->fpath);
-        rc = DYAD_RC_BADFIO;
-    } else {
-        *file_len = (size_t)read_len;
+    if (ctx->dtl_handle->mode == DYAD_DTL_UCX) {
```
JaeseungYeom, line 779: This is the UCX guard added to prevent the FLUX RPC flow from reaching.

src/dyad/client/dyad_client.c
```diff
@@ -772,17 +1129,24 @@ dyad_rc_t dyad_consume (dyad_ctx_t *restrict ctx, const char *restrict fname)
...
     } else {
-        if (file_size <= 0) {
+        // When use_fs_locks is false, filesystem locking is unavailable
+        // (e.g. DYAD_HAS_STD_FSTREAM_FD is not defined for C++ streams),
+        // so we cannot rely on file size alone to determine if the file is
+        // fully written as producer cannot lock the file it is still writting.
+        // In that case, always fetch metadata from KVS to ensure correctness,
+        // as in the shared storage path. For the C GOTCHA wrapper path,
+        // use_fs_locks is irrelevant and should always be true.
+        if (!ctx->use_fs_locks || file_size <= 0) {
```
JaeseungYeom, line 1149: This is to support the fallback case of C++ stream wrapper that does not have the file locking capability.
Fortunately, we always had it from the compilers we have on LC.
To place a lock on the file managed by a stream, we rely on an internal function that returns file descriptor of basic_filebuf. However, this is compiler implementation specific and not part of the standard. We do run cmake test at configure time to see it is available. If so, the cmake will define the macro `DYAD_HAS_STD_FSTREAM_FD` and set `ctx->use_fs_locks` to `true`. Otherwise, false in `dyad_stream_core::init()` If DYAD_HAS_STD_FSTREAM_FD is not defined, the c++ producers (ofstream/fstream wrappers) have no means to place a lock on file descriptor. That means, consumer can read a partially written files by local producers. That is why we force dyad_fetch_metadata() here in dyad_consume() to synchronize local producers and consumers via KVS. However, C-wrappers are always able to place locks on file descriptors. For C interception wrappers, `use_fs_lock` should always be true, which is why it is set in `dyad_wrapper_init()` (see wrapper.c)

src/dyad/wrapper/wrapper.c
```diff
@@ -134,6 +156,10 @@ void dyad_wrapper_init (void)
     DYAD_C_FUNCTION_START ();
     dyad_ctx_init (DYAD_COMM_RECV, NULL);
     ctx = ctx_mutable = dyad_ctx_get ();
+    // See dyad_consume () in dyad_client.c
+    // TODO: In case that the wrapper and c++ stream wrapper class co-exist
+    // this variable should be context dependent.
+    ctx_mutable->use_fs_locks = true;
```
JaeseungYeom, line 159: I left comment on this in client.c, specifically in dyad_consume()

include/dyad/stream/dyad_stream_api.hpp
```diff
@@ -217,32 +379,80 @@ void unlock_fstream (std::basic_fstream<_CharT, _Traits>& os, const dyad_stream_
...
+ * When @c DYAD_HAS_STD_FSTREAM_FD is not defined, all macros expand to
+ * nothing. Filesystem locking is silently disabled and synchronization
+ * falls back to KVS-based coordination only, which may perform worse for
+ * co-located producers and consumers on node-local storage since the
+ * producer cannot lock the file it is writing.
+ *
+ * @see dyad_stream_core::init(const dyad_params&) where @c use_fs_locks
+ *      is set accordingly.
+ */
+#if defined(DYAD_HAS_STD_FSTREAM_FD)
```
JaeseungYeom, line 414: separated this from a large block so that I can leave comment block isolated to this macro set.

src/dyad/stream/dyad_stream_core.cpp
```diff
@@ -283,7 +286,7 @@ int dyad_stream_core::file_lock_shared (int fd) const
 {
     struct flock shared_flock;
 
-    dyad_rc_t rc = dyad_excl_flock (m_ctx, fd, &shared_flock);
+    dyad_rc_t rc = dyad_shared_flock (m_ctx, fd, &shared_flock);
```
JaeseungYeom, line 289: This is obviously an error.

PR #178, opened 2026-06-05
Optimizes the Margo DTL's memory usage by matching the behavior of the UCX DTL

ilumsden:
I was using DYAD on Tuolumne recently, and I found two issues with DYAD.

First, I found that DYAD would crash whenever trying to send/recv data with a Mercury NA protocol string of `ofi+cxi` (i.e., what should be used for Slingshot). This was due to an oversight in the `dyad_dtl_margo_rpc_pack` function. That function assumes that the network address will fit into a statically sized 128 byte buffer. That assumption is invalid on Tuo because the Slingshot network normally uses network addresses larger than 128 bytes.

This PR fixes this first issue by updating `dyad_dtl_margo_rpc_pack` to call `margo_addr_to_string` twice. The first call is used to get the actual length of the network address. Then, after a `malloc`, the second call is used to actually obtain the network address in full.

Second, I found that DYAD was performing much worse than expected. When reviewing the code, I found that the main issue was in how the Margo DTL was managing memory.

By allocating a buffer and calling `margo_bulk_create` for every send/recv, the DTL was essentially incurring the cost of memory allocation + memory pinning on the NIC repeatedly throughout a run. To add onto that, when I started using DYAD on Tuo, the Margo bulk objects were never being freed, resulting in massive memory leaks. Those memory leaks were recently fixed by other PRs, but the underlying alloc + pin cost still exists.

To fix this issue, this PR mimics the memory reuse scheme used by the UCX DTL. It now creates a single allocation and calls `margo_bulk_create` once during initialization, and it frees and unpins the memory during finalization. On the sender side, the pinned memory buffer is returned to the broker module via `dyad_dtl_margo_get_buffer` to ensure the module reads directly from local storage into the pinned memory. Similarly, on the receiver side, the pinned memory is passed directly into the `margo_bulk_transfer` call to avoid excessive allocations. The `dyad_dtl_margo_recv` function then copies from that pinned memory into a new allocation to return to the DYAD client. This PR also adds C11's `stdatomic.h` to help the `dyad_dtl_margo_recv` and `data_ready_rpc` synchronize on data availability.

This PR also makes the max transfer size for the UCX and Margo DTL backends configurable via the new CMake cache variable `DYAD_DTL_MAX_TRANSFER_SIZE`. The value of this variable defaults to 4294967296 (i.e., 4 GiB) which matches the existing behavior of the UCX DTL.

I tested these changes on Tuo. They result in much better performance. In fact, with these changes, DYAD can achieve results comparable to a data transfer approach using `MPI_Isend` and `MPI_Irecv` in certain workflow situations.


src/dyad/dtl/margo_dtl.c
```diff
@@ -598,14 +652,21 @@ dyad_rc_t dyad_dtl_margo_recv (const dyad_ctx_t *ctx, void **buf, size_t *buflen
...
-    // margo_handle->recv_buffer is allocated in data_ready_rpc()
-    free (margo_handle->recv_buffer);
+    // margo_handle->recv_buffer is never actually allocated itself when in RECV mode.
+    // Instead, data_ready_rpc() will set it to point to the margo_handle->bulk_buffer, which is
+    // reused between different receive operations. So, we just need to set recv_buffer back to
+    // NULL to "release" it.
     margo_handle->recv_buffer = NULL;
     margo_handle->recv_len = 0;
-    margo_handle->recv_ready = 0;
+    atomic_store_explicit (&margo_handle->recv_ready, false, memory_order_relaxed);
```
JaeseungYeom, line 667: - I am curious about this. It is good to know that this variable needs to be protected. However, was there incident?
- Reusing buffer would improve performance. However, that also means that there will be no concurrent transfer.
- `stdatomic` seems to be optional with c11. We either bump up the minimum C standard required if the protection is necessary. Otherwise, do test as in `cmake/tests` at cmake configure time and propagate the choice via `dyad_config.h`

src/dyad/dtl/CMakeLists.txt
```diff
@@ -12,12 +12,12 @@ set(FLUX_PRIVATE_HEADERS ${CMAKE_CURRENT_SOURCE_DIR}/flux_dtl.h)
 set(FLUX_PUBLIC_HEADERS)
 
 # UCX implementation for DTL
-set(UCX_DTL_SRC ${CMAKE_CURRENT_SOURCE_DIR}/ucx_dtl.c ${CMAKE_CURRENT_SOURCE_DIR}/ucx_ep_cache.cpp)
+set(UCX_DTL_SRC ${CMAKE_CURRENT_BINARY_DIR}/ucx_dtl.c ${CMAKE_CURRENT_SOURCE_DIR}/ucx_ep_cache.cpp)
```
JaeseungYeom, line 15: Instead of doing this, MAX_TRANSFER_SIZE as an env variable.
And parse size characters at the end.
We wouldn't need to change CMakeLists.txt and make it needlessly static.

src/dyad/dtl/margo_dtl.c
```diff
@@ -23,6 +23,10 @@
 #include <mercury.h>
 #include <mercury_macros.h>
 
+// clang-format off
+#define MARGO_MAX_TRANSFER_SIZE (@DYAD_DTL_MAX_TRANSFER_SIZE@)
```
JaeseungYeom, line 27: This sould be converted into the env variable DYAD_DTL_MAX_TX_SIZE_ENV "DYAD_DTL_MAX_TX_SIZE"

src/dyad/dtl/margo_dtl.c
```diff
@@ -112,26 +114,32 @@ static void data_ready_rpc (hg_handle_t h)
...
     margo_handle->recv_len = (size_t)in.n;
-    margo_handle->recv_buffer = malloc (margo_handle->recv_len);
-
-    ret = margo_bulk_create (mid,
-                             1,
-                             (void **)&margo_handle->recv_buffer,
-                             &margo_handle->recv_len,
-                             HG_BULK_WRITE_ONLY,
-                             &local_bulk);
-    assert (ret == HG_SUCCESS);
```
JaeseungYeom, line 123: We should actually get rid of all the asserts in this file and handle it properly.
But that is for another PR.

src/dyad/dtl/margo_dtl.c
```diff
@@ -95,13 +99,11 @@ static void data_ready_rpc (hg_handle_t h)
     hg_return_t ret;
     margo_rpc_in_t in;
     margo_rpc_out_t out;
-    hg_bulk_t local_bulk;
     // clang-format off
     (void) ret;
     // clang-format on
 
     margo_instance_id mid = margo_hg_handle_get_instance (h);
-    margo_set_log_level (mid, MARGO_LOG_INFO);
```
JaeseungYeom, line 104: #if define(DYAD_LOGGER_LEVEL_INFO)
    margo_set_log_level (mid, MARGO_LOG_INFO);
#endif

src/dyad/dtl/margo_dtl.c
```diff
@@ -154,21 +162,19 @@ dyad_rc_t dyad_dtl_margo_get_buffer (const dyad_ctx_t *ctx, size_t data_size, vo
     DYAD_C_FUNCTION_START ();
     dyad_rc_t rc = DYAD_RC_OK;
 
-    if (data_buf == NULL || *data_buf != NULL) {
+    dyad_dtl_margo_t *margo_handle = ctx->dtl_handle->private_dtl.margo_dtl_handle;
+
+    if (data_size > margo_handle->bulk_buffer_size) {
```
JaeseungYeom, line 167: In both UCX and Margo, if this happens, we should simply reallocate the buffer with a bigger size.

src/dyad/dtl/margo_dtl.c
```diff
@@ -179,11 +185,11 @@ dyad_rc_t dyad_dtl_margo_return_buffer (const dyad_ctx_t *ctx, void **data_buf)
 {
     DYAD_C_FUNCTION_START ();
     dyad_rc_t rc = DYAD_RC_OK;
-    if (data_buf == NULL || *data_buf == NULL) {
```
JaeseungYeom, line 182: I see that static buffer is already checked but it is still safer to check especially if we are going to reallocate.
In UCX case, you had a different problem, an unsolved mystery. That is something to revisit as well.

src/dyad/dtl/margo_dtl.c
```diff
@@ -281,6 +287,9 @@ dyad_rc_t dyad_dtl_margo_init (const dyad_ctx_t *ctx,
 
     // dyad_rc_t rc = DYAD_RC_OK;
     dyad_dtl_margo_t *margo_handle = NULL;
+    dyad_rc_t rc = 0;
```
JaeseungYeom, line 290: Initialize it with a return code macro.

src/dyad/dtl/margo_dtl.h
```diff
@@ -9,6 +9,7 @@
 
 #include <dyad/dtl/dyad_dtl_api.h>
 #include <margo.h>
+#include <stdatomic.h>
```
JaeseungYeom, line 12: This is guaranteed to be available in C11.

src/dyad/service/flux_module/dyad.c
```diff
@@ -141,7 +141,10 @@ void dyad_mod_fini (void)
 static void freectx (void *arg)
 {
     dyad_mod_ctx_t *mod_ctx = (dyad_mod_ctx_t *)arg;
-    flux_msg_handler_delvec (mod_ctx->handlers);
+    if (mod_ctx->handlers != NULL) {
```
JaeseungYeom, line 144: Good!

CMakeLists.txt
```diff
@@ -106,7 +106,7 @@ if (DYAD_ENABLE_MARGO_DATA)
     set (DYAD_ENABLE_MARGO_DTL 1)
 endif()
 
-
+set(DYAD_DTL_MAX_TRANSFER_SIZE "4294967296" CACHE STRING "Maximum transfer size supported by DYAD in bytes")
```
JaeseungYeom, line 109: We want this be rather runtime configurable.
CMake can set the deafult value.
DYAD_DTL_MAX_TX_SIZE_DEFAULT, which should be written to dyad_config.h
This should be commonly used for both UCX and MARGO, client and server.

JaeseungYeom (changes requested):
I have not finished reviewing margo_dtl.c yet but I will come back later.

CMakeLists.txt
JaeseungYeom: This one is from shuffle test.
bool parse_size (const char* str, size_t& out)
{
    char* end;
    long long val = strtoll (str, &end, 10);
    if (val <= 0 || end == str)
        return false;


    switch (*end) {
        case 'K':
        case 'k':
            out = static_cast<size_t> (val) * 1024ULL;
            break;
        case 'M':
        case 'm':
            out = static_cast<size_t> (val) * 1024ULL * 1024ULL;
            break;
        case 'G':
        case 'g':
            out = static_cast<size_t> (val) * 1024ULL * 1024ULL * 1024ULL;
            break;
        case '\0':
            out = static_cast<size_t> (val);
            break;
        default:
            return false;  // unrecognized suffix
    }
    return true;
}

JaeseungYeom:
Minimum to merge is to make the transfer size parameter environment variable controllable.

src/dyad/dtl/margo_dtl.c
```diff
@@ -408,13 +437,41 @@ dyad_rc_t dyad_dtl_margo_rpc_pack (const dyad_ctx_t *ctx,
...
+    hg_rc =
+        margo_addr_to_string (margo_handle->mid, NULL, &addr_str_size, margo_handle->local_addr);
+    if (hg_rc != HG_SUCCESS) {
+        DYAD_LOG_ERROR (ctx, "[MARGO DTL] margo_addr_to_string size query failed!");
+        rc = DYAD_RC_SYSFAIL;
+        goto dtl_margo_rpc_pack_region_finish;
+    }
+    addr_str = malloc (addr_str_size);
+    if (addr_str == NULL) {
+        rc = DYAD_RC_SYSFAIL;
```
JaeseungYeom, line 463: TODO: Need to define a RC macro for this as well as other cases where assert is called.

src/dyad/dtl/margo_dtl.c
```diff
@@ -634,12 +695,23 @@ dyad_rc_t dyad_dtl_margo_finalize (const dyad_ctx_t *ctx)
...
+        margo_handle->bulk_handle = HG_BULK_NULL;
+    }
+
     if (margo_handle->mid != MARGO_INSTANCE_NULL) {
         margo_addr_free (margo_handle->mid, margo_handle->local_addr);
         if (margo_handle->remote_addr != NULL)
             margo_addr_free (margo_handle->mid, margo_handle->remote_addr);
         margo_finalize (margo_handle->mid);
     }
+
```
JaeseungYeom, line 709: Move this up before `if (margo_handle->mid != MARGO_INSTANCE_NULL) {`

src/dyad/dtl/margo_dtl.c
```diff
@@ -634,12 +695,23 @@ dyad_rc_t dyad_dtl_margo_finalize (const dyad_ctx_t *ctx)
...
 
+    if (margo_handle->bulk_handle != HG_BULK_NULL) {
+        margo_bulk_free (margo_handle->bulk_handle);
+        margo_handle->bulk_handle = HG_BULK_NULL;
+    }
+
     if (margo_handle->mid != MARGO_INSTANCE_NULL) {
         margo_addr_free (margo_handle->mid, margo_handle->local_addr);
         if (margo_handle->remote_addr != NULL)
             margo_addr_free (margo_handle->mid, margo_handle->remote_addr);
```
JaeseungYeom, line 706: set margo_handle->remote_addr = HG_ADDR_NULL after margo_addr_free()

src/dyad/dtl/margo_dtl.c
```diff
@@ -634,12 +695,23 @@ dyad_rc_t dyad_dtl_margo_finalize (const dyad_ctx_t *ctx)
 
     margo_handle = ctx->dtl_handle->private_dtl.margo_dtl_handle;
 
+    if (margo_handle->bulk_handle != HG_BULK_NULL) {
+        margo_bulk_free (margo_handle->bulk_handle);
+        margo_handle->bulk_handle = HG_BULK_NULL;
+    }
+
     if (margo_handle->mid != MARGO_INSTANCE_NULL) {
         margo_addr_free (margo_handle->mid, margo_handle->local_addr);
```
JaeseungYeom, line 704: margo_handle->local_addr = HG_ADDR_NULL

src/dyad/dtl/margo_dtl.c
```diff
@@ -634,12 +695,23 @@ dyad_rc_t dyad_dtl_margo_finalize (const dyad_ctx_t *ctx)
...
         if (margo_handle->remote_addr != NULL)
```
JaeseungYeom, line 705: margo_handle->remote_addr != HG_ADDR_NULL

src/dyad/dtl/margo_dtl.c
JaeseungYeom, line 401: margo_handle->remote_addr = HG_ADDR_NULL

PR #183, opened 2026-07-17
Add cache eviction, byte-range fetch, and worker-pool concurrency

wangvsa:
## Summary

Adds opt-in node-local cache eviction and origin-backed byte-range fetch to
DYAD, makes the Flux module's fetch-RPC path concurrent (fixing a real
single-reactor-thread serialization bottleneck), and fixes two correctness
bugs found while hardening the Margo DTL backend and pydyad under load.

- **Node-local cache eviction + byte-range fetch** (`dyad_consume_range()`):
  pluggable LRU/FIFO eviction for DYAD-managed directories
  (`DYAD_CACHE_CAPACITY`/`_POLICY`/`_LOW_WATERMARK`/`_GRACE_PERIOD`), plus a
  byte-range consume API that fetches directly into memory with an optional
  PFS/origin fallback (`DYAD_PATH_ORIGIN`) that lazily fills only the
  requested span of a managed file on first touch. Implemented for the
  FLUX_RPC and Margo DTL backends.
- **Concurrent fetch-RPC servicing**: the module's single Flux reactor
  thread previously serviced every `dyad.fetch_range` request fully
  synchronously — including blocking file I/O and, for Margo, the RDMA send
  itself — so concurrent requests from independent ranks/nodes queued
  behind whichever one happened to arrive first. Adds a fixed-size
  worker-thread pool (`DYAD_FETCH_WORKER_THREADS`, default 8) via a new
  DTL vtable detach/send-detached extension, implemented for FLUX_RPC and
  Margo (UCX keeps the original synchronous path).
- **Margo DTL fixes**: a per-request RDMA memory-registration leak
  (`margo_bulk_create()` never freed on the success path, exhausting the
  CXI NIC's finite MR table under sustained load) and a per-consumer
  address cache (repeated `margo_addr_lookup()`/`margo_addr_free()` churn
  was found to exhaust/corrupt Mercury's CXI provider state under
  sustained request volume).
- **pydyad fix**: `Dyad.finalize()` never reset `self.initialized`, so its
  guard against double-finalization never actually fired — a second call
  (e.g. an explicit `finalize()` followed by `__del__()` at interpreter
  exit) could call `dyad_finalize()` twice on already-freed C-side state.

## Test plan

- [x] Each commit builds cleanly on its own (verified incrementally while
      splitting into this commit sequence)
- [x] `tests/pydyad_spsc_range`/`tests/pydyad_spsc_lazy_range` byte-range
      correctness (bytes match origin) under both `DYAD_DTL_MODE=FLUX_RPC`
      and `DYAD_DTL_MODE=MARGO`
- [x] Real PyTorch training workload (PECAN/EGNN) at 1–4 node / 4–16 GPU
      scale, both DTL backends, including 1000+-iteration runs with the
      worker pool and Margo fixes active — zero errors
- [ ] New unit tests for the cache-eviction policies
      (`tests/unit/cache/cache_functions.cpp`) — included, not yet run in
      CI here

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JaeseungYeom (approved):
Looks good to me.

